PR #1, opened 2020-02-06
Initial commit

justinboswell:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


.github/actions/release-tag/index.js
```diff
@@ -0,0 +1,26 @@
+const core = require('@actions/core');
```
justinboswell: This is new

.github/actions/release-tag/packed/index.js
```diff
@@ -0,0 +1,400 @@
+module.exports =
```
justinboswell: this is a build artifact, ignore, it's based on package.json + index.js

.github/workflows/docker-images.yml
```diff
@@ -0,0 +1,289 @@
+name: Docker Images
```
justinboswell: This builds all the docker images for the PR branch or release

builder/__main__.py
```diff
@@ -0,0 +1,1500 @@
+# Copyright 2010-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
justinboswell: This used to be builder.py

graebm: I have never seen a `__main__.py` file before
this might be python3-only?
I don't personally care if it breaks python2, we could always rename it later and its usage would stay the same

.github/workflows/docker-images.yml
```diff
@@ -0,0 +1,289 @@
...
+      # re-enable this when requirements.txt has actual dependencies
+      # - name: Bundle dependencies
+      #   run: |
+      #     python3 -m pip install -r requirements.txt --target builder
+      #     rm -rf builder/*.dist-info
+
+      - name: Package builder
+        run: |
+          mkdir -p build
+          python3 -m zipapp --python="/usr/bin/env python3" --output=build/builder builder
```
graebm: ok, we're definitely python3-only now
heck yeah

PR #7, opened 2020-02-14
Fixes for issues encountered converting Java CRT

justinboswell:


builder/__main__.py
```diff
@@ -38,7 +37,7 @@
 ########################################################################################################################
 
 def run_action(action, env):
-
+    config = env.config
     # Set build environment from config
     env.shell.pushenv()
     for var, value in config.get('build_env', {}).items():
```
graebm: This was crashing before, right?

It wasn't picking up some other `config` defined somewhere else? If this was the case, let's change a variable name somewhere so this mistake isn't so easy to make.

but probably this was just a crash

builder/actions/git.py
```diff
@@ -32,15 +32,16 @@ def run(self, env):
 
         sh = env.shell
 
-        sh.exec("git", "clone", self.project.url, self.path, always=True)
+        sh.exec("git", "clone", self.project.url,
+                self.path, always=True, retries=3)
```
graebm: retries. fucking neat

builder/__main__.py
justinboswell: Yeah, it was picking up the one at global scope from __main__.  I got rid of that.

builder/actions/git.py
justinboswell: I've seen every git command and apt command fail while testing. This is a necessity, I'm tired of "Well just push or run it again"

graebm: I was serious.
Retries ARE neat.
It was cool to see you already had support for that

PR #11, opened 2020-02-25
Fixed libcrypto in docker containers, misc C++ fixes

justinboswell:

### DIFF
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   cleanup:
+    name: Clean up previous branch jobs
     runs-on: ubuntu-latest
     steps:
     - name: Clean up previous branch jobs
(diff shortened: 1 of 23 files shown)


.github/workflows/sanity-test.yml
```diff
@@ -50,4 +59,23 @@ jobs:
       - name: Sanity Test Build (Windows)
         if: matrix.host == 'windows-latest'
         run: |
-          python -m builder build --project tests
\ No newline at end of file
+          python -m builder build --project tests
+
+  # downstream_test:
```
justinboswell: I'll turn these on once all these projects are working

builder/actions/cmake.py
```diff
@@ -68,16 +68,12 @@ def build_project(project, build_tests=False):
             # Set compiler flags
             compiler_flags = []
             if toolchain.compiler != 'default':
-                compiler_path = toolchain.compiler_path(env)
-                if compiler_path:
-                    for opt in ['c', 'cxx']:
+                c_path = toolchain.compiler_path(env)
```
justinboswell: these should never have been the same path

PR #14, opened 2020-03-03
Fixes found porting java/cpp/python

justinboswell:
Major items:
* Toolchain no longer depends on Env, so Toolchain functions can be used at any time, anywhere
* Greatly simplified arg parsing
* Added environment injection for cross-compile targets
* Fixed a bug that was causing configs to mashed into each other and becoming a mess
* Got rid of dependency on tar/curl
* Fixed a problem where GC would destroy dynamically loaded classes before we got to use them, resulting in a truly insane debugging situation
* Made sure all downloads from CloudFront use a param on the end to avoid the cache

### DIFF
--- a/builder/scripts.py
+++ b/builder/scripts.py
@@ -49,9 +49,6 @@ class Scripts(object):
     # Must cache all classes with a reference here, or the GC will murder them
     all_classes = set()
 
-    def __init__(self):
-        Scripts.all_classes = _get_all_dynamic_classes()
-
     @staticmethod
     def load(path='.'):
         """ Loads all scripts from ${path}/.builder/**/*.py to make their classes available """
--- a/builder/scripts.py
+++ b/builder/scripts.py
@@ -16,15 +16,41 @@ import glob
 import os
 import sys
 
-from action import Action
+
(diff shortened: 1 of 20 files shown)


builder/scripts.py
```diff
@@ -16,15 +16,41 @@
...
 class Scripts(object):
     """ Manages loading, context, and running of per-project scripts """
-    all_actions = set()
+
+    # Must cache all classes with a reference here, or the GC will murder them
+    all_classes = set()
 
     def __init__(self):
-        Scripts.all_actions = set(Action.__subclasses__())
+        Scripts.all_classes = _get_all_dynamic_classes()
```
graebm: Confused about static/non-static here.
`Scripts` is class with all `static` functions
`Scripts.all_classes` is class variable (static, essentiall)
But `Scripts.all_classes` gets populated in the instance constructor

When does the `Scripts.__init__()` ever get run? Is there a Scripts instance somewhere?

justinboswell: Oh man you're right, it used to get instanced when we had that global builder instance. I'll fix this up.

PR #31, opened 2020-03-17
More fixes

justinboswell:
* Fixed case where deps could be installed multiple times because multiple copies of the import were instantiated
* Fixed project directory overrides

### DIFF
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -234,10 +234,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Get release tag
-        uses: ./.github/actions/release-tag
-        id: release
-        with:
-          output: tag
+      uses: ./.github/actions/release-tag
+      id: release
+      with:
+        output: tag
 
     - name: Build ${{ env.PACKAGE_NAME }}
       run: |
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -233,6 +233,9 @@ jobs:
     needs: package
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout Source
+      uses: actions/checkout@v1
+      
     - name: Get release tag
       uses: ./.github/actions/release-tag
       id: release
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -235,7 +235,7 @@ jobs:
     steps:
     - name: Checkout Source
       uses: actions/checkout@v1
-      
+
     - name: Get release tag
       uses: ./.github/actions/release-tag
       id: release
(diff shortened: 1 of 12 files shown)


builder/data.py
```diff
@@ -37,8 +37,8 @@ class PKG_TOOLS(Enum):
 
     # where the cmake binaries should be stored, and dependencies installed
     'build_dir': 'build',
-    'deps_dir': 'build/deps',
-    'install_dir': 'build/install',
+    'deps_dir': '{build_dir}/deps',
```
justinboswell: This allows the default deps_dir and install_dir to be based on the build_dir if it is specified in config, but they are not.

builder/env.py
```diff
@@ -96,29 +96,28 @@ def __setitem__(self, item, value):
 
         # Once initialized, switch to the source dir before running actions
         self.source_dir = os.path.abspath(self.project.path)
+        self.variables['source_dir'] = self.source_dir
```
justinboswell: This order of operations allows the config dirs to depend on each other

builder/imports/jdk.py
```diff
@@ -0,0 +1,127 @@
+# Copyright 2010-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
justinboswell: We're going to need this in more than just the Java CRT (think jruby, or any other JVM language)

builder/imports/s2n.py
```diff
@@ -15,24 +15,25 @@
...
     'test_steps': [],
-    'cmake_args': {
-        '-DS2N_NO_PQ_ASM=ON',
-    },
+    'build_tests': False,
 }
 
 
-class S2N(Project, Import):
+class S2NImport(Import):
```
justinboswell: The previous multiple inheritance was a terribly clever and awful idea

builder/project.py
```diff
@@ -547,22 +562,20 @@ def _project_from_path(path='.', name_hint=None):
...
 
-                if not project_config.get('name', None):
-                    project_config['name'] = name_hint if name_hint else os.path.dirname(
-                        os.getcwd())
-                print('    Found project: {} at {}'.format(
-                    project_config['name'], path))
-                project = Project._create_project(
-                    **project_config, path=path)
-                return Project._cache_project(project)
+                if name_hint == None or project_config.get('name', None) == name_hint:
```
justinboswell: Previously, this and the path assignment below were leading every project to just return the root project as its path, creating the infinite recursion

builder/project.py
```diff
@@ -587,17 +600,17 @@ def find_project(name, hints=[]):
...
                 project = Project._project_from_path(search_dir, name)
 
                 if project:
                     return project
 
                 # might be a project without a config
-                if looks_like_code(search_dir):
-                    print(
-                        ('    Found source code only project at {}'.format(search_dir)))
+                if dir_matches_name and looks_like_code(search_dir):
```
justinboswell: This was equally at fault for the recursion

.github/workflows/sanity-test.yml
```diff
@@ -215,6 +219,37 @@ jobs:
...
+    - name: Get release tag
+      uses: ./.github/actions/release-tag
+      id: release
+      with:
+        output: tag
+
+    - name: Build ${{ env.PACKAGE_NAME }}
+      run: |
+        echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u awslabs --password-stdin
+        # TODO: change this to latest, let the version be dictated by --version below
```
graebm: is this a "todo before merge" or "todo someday"?

justinboswell: Before merge

PR #34, opened 2020-03-24
Implemented publish to S3, mirror for JDK

justinboswell:
The goal here is to reduce our reliance on public infrastructure where we can, and to reduce downloads where possible. We should be able to bake the cache into our docker images, or mount it as a tarball volume, for instance.

This is in preparation for the NDK, which is a couple gigs that I don't want to have to download everywhere all the time.

### DIFF
--- a/builder/fetch.py
+++ b/builder/fetch.py
@@ -19,6 +19,7 @@ import tempfile
 import time
 import tarfile
 import zipfile
+from urllib.parse import urlparse
 from urllib.request import urlretrieve, urlopen
 
 try:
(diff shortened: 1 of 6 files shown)


builder/fetch.py
```diff
@@ -31,8 +32,14 @@
 except:
     msvcrt = None
 
+from util import run_command
 
-MANIFEST_URL = 'https://d19elf31gohf1l.cloudfront.net/_binaries/MANIFEST'
+FETCH_URL = 'https://d19elf31gohf1l.cloudfront.net/_binaries'
```
bretambrose: How stable is this?

builder/fetch.py
```diff
@@ -237,3 +256,30 @@ def fetch_script(url, script_path):
 
     print('Applying exec permissions to {}'.format(script_path))
     os.chmod(script_path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
+
+
+def publish_package(name, package_path):
```
bretambrose: minor, but "publish" makes me think maven vs. copy-to-s3

builder/fetch.py
justinboswell: It won't change unless we de/re-provision cloudfront AFAIK. And maybe not even then.

builder/fetch.py
justinboswell: What would you prefer? I can update it.

bretambrose: copy_package_to_s3 maybe?

justinboswell: I feel like that leaks the abstraction a bit. I would like to keep the storage end of this obfuscated, in case it changes to a package manager or something in the future.

justinboswell: If we want to change it later, I'm happy to. It's an easy refactor.

PR #45, opened 2020-06-26
Added support for finding working directory via codebuild

justinboswell:


.github/docker-images/entrypoint.sh
```diff
@@ -27,5 +27,12 @@ if [ -x /opt/python/cp37-cp37m/bin/python ]; then
     ln -s /opt/python/cp37-cp37m/bin/python /usr/local/bin/python3
 fi
 
-cd $GITHUB_WORKSPACE
+# Figure out where to work based on environment, default to .
```
DavidOgunsAWS, line 30: This value resolution looks useful to put in a variable. It's future speculation and should wait until that point though.

PR #52, opened 2020-09-21
`libcrypto_path` variable is now an absolute path.

graebm:
aws-crt-ruby does not build from the repo root, so a relative path wasn't working out


graebm:
relative path was needed for cross-compile

PR #59, opened 2020-10-01
Added additional helpful docs to builder README

justinboswell:

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ See builder/data.py for more info/defaults/possible values.
     // * version - the compiler version
     // * target - the target OS class (linux/windows/macos/android)
     // * arch - the target arch
-    // * cwd - the current working directory (affected by --build-dir)
+    // * cwd - the current working directory (affected by --build-dir argument)
     // * source_dir - The source directory for the project being built
     // * build_dir - The directory where intermediate build artifacts will be generated (defaults to "{source_dir}/build")
     // * deps_dir - The root directory where dependencies will be installed (defaults to "{build_dir}/deps")
--- a/README.md
+++ b/README.md
@@ -139,19 +139,24 @@ See builder/data.py for more info/defaults/possible values.
     // value can be overridden from within these override sections, see below for examples.
 
     // Configuration differences per host (the machine/image the build runs on)
+    // Any host not specified will be built with default values from the rest of the config
     "hosts": {
         "linux": {}, // includes all flavors of linux below
         "ubuntu": {},
         "debian": {},
         "al2": {},
-        "al2012": {},
+        "al2012": {
+            "enabled": false // example: disable building on AL2012
+        },
         "alpine": {},
         "raspbian": {},
         "manylinux": {},
         "macos": {},
         "windows": {}
     },
+
     // Configuration differences per target platform (the machine being built for)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,27 +113,34 @@ NOTE: Any key can be prefixed with a ```!``` to overwrite the config value, rath
...
     "imports": [
         "s2n"
     ],
-    // Configuration differences per host
+
+    // Per-environment overrides
+    // Overrides are applied per host, per target/architecture, and per compiler/version. Any top-level config 
+    // value can be overridden from within these override sections, see below for examples.
+
+    // Configuration differences per host (the machine/image the build runs on)
```
kneekey23, line 141: something about whether or not you need to include each host you want to test on or only including the ones you don't want with "enabled": false

kneekey23 (approved):
LGTM

README.md
```diff
@@ -61,13 +61,39 @@ that instead. There are a few external dependencies (s2n and libcrypto, for inst
...
 
 #### Detailed config:
-NOTE: Any key can be prefixed with a ```!``` to overwrite the config value, rather than to add to it.
+NOTE: Any key can be prefixed with a ```!``` to overwrite the config value, rather than to add to it. 
+      Any key can be prefixed with a ```+``` to force a value to be added to an array.
 
-```json
+See builder/data.py for more info/defaults/possible values.
+
+```jsonc
```
graebm, line 69: thank you 1000 for fixing this. the red highlights burned my eyes

README.md
```diff
@@ -188,6 +224,14 @@ of common shell operations (cd, cwd, pushd, popd, setenv, getenv, pushenv, popen
 arbitrary commands.
 
 ## Developing on builder
+
+### Debugging
+When debugging builder locally, use whatever python debugger you wish. You can also feed it the following command line arguments to ease the
+debugging experience:
+* --skip-install - don't install packages, assume they're already there
+* --build-dir=/path/to/other/git/repo - will jump to this directory before starting execution, helpful for debugging another project's
```
graebm, line 239: I thought it was `--project` ?

justinboswell: --project can take a path, but --build-dir will look in that directory for additional config/scripts as well (it's really only useful for setting the cwd during development, normally this app would get run from a source directory, or in an empty directory and pull its own source)

README.md
```diff
@@ -61,64 +61,121 @@ that instead. There are a few external dependencies (s2n and libcrypto, for inst
...
+    // per host/os/target/architecture/compiler/version (see below)
+    // The following variables are pre-defined by builder internally:
+    // * host - the host OS
+    // * compiler - the compiler that will be used if project needs a compiler
+    // * version - the compiler version
+    // * target - the target OS class (linux/windows/macos/android)
+    // * arch - the target arch
+    // * cwd - the current working directory (affected by --build-dir)
+    // * source_dir - The source directory for the project being built
+    // * build_dir - The directory where intermediate build artifacts will be generated
```
kneekey23, line 89: if this is the root, don't include maybe? unless you feel its self explanatory

justinboswell: I added some clarifying defaults, does that help?

graebm (approved):

PR #71, opened 2020-11-12
Update how node is installed on windows CI builds so that it uses 10.16

bretambrose:
Changes to dependencies resulted in node not getting installed properly and when it did get installed it was using the minimum platform (10.0.0) rather than what we need (10.16+)

### DIFF
--- a/builder/imports/nodejs.py
+++ b/builder/imports/nodejs.py
@@ -29,7 +29,10 @@ class NodeJS(Import):
             },
             **kwargs)
         self.url = 'https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh'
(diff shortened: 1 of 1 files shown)


builder/imports/nodejs.py
```diff
@@ -29,7 +29,10 @@ def __init__(self, **kwargs):
             },
             **kwargs)
         self.url = 'https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh'
-        self.version = '10'
+        if current_os() == 'windows':
```
justinboswell, line 32: `# windows nvm doesn't install the latest revision, unix does`

justinboswell (approved):

PR #76, opened 2020-11-18
Fixed ios armv8 aliases, fixed cross compile logic for ios

justinboswell:

### DIFF
--- a/builder/toolchain.py
+++ b/builder/toolchain.py
@@ -76,8 +76,8 @@ def _msvc_versions():
 
 def _is_cross_compile(target_os, target_arch):
     # Mac compiling for anything that isn't iOS or itself
-    if current_os() == 'macos' and target_os not in ["macos", "ios"]:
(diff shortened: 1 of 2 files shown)


builder/toolchain.py
```diff
@@ -76,8 +76,8 @@ def _msvc_versions():
 
 def _is_cross_compile(target_os, target_arch):
     # Mac compiling for anything that isn't iOS or itself
-    if current_os() == 'macos' and target_os not in ["macos", "ios"]:
-        return True
+    if current_os() == 'macos' and target_os in ["macos", "ios"]:
+        return False
```
kneekey23, line 80: this is much more readable :) +1

kneekey23 (approved):
LGTM

builder/toolchain.py
justinboswell: It also has the benefit of being correct!

Issue #88, opened 2020-12-22
released python source artifacts unusable


aajtodd:
Was trying to get this setup for local use to explore what I need to do to integrate CI into another CRT language binding repo.

I figured I'd dowload the latest tagged release [0.7.6](https://github.com/awslabs/aws-crt-builder/releases/tag/v0.7.6) and install it to a virtualenv.

I eventually got it working only after hacking the `setup.py` script and a few other things. I'll just layout what I see as some areas for improvement.

1. `setup.py` [relies](https://github.com/awslabs/aws-crt-builder/blob/main/setup.py#L7-L11) on git for the version information which isn't available in a released artifact. This should be static once released (it currently throws an exception). You can probably just change this to spit out a version file and use that when present (and include it in the artifact).
2. The `setup.py` script fails to install locally for me on osx Python 3.7.3. The error I get is the following
```
error: file '/private/var/folders/qc/cf9667wn5sbbgknq1h25_gfcxjjs4r/T/pip-req-build-e4sdc7h1/builder/*.py' does not exist
```
This stems from how [scripts](https://github.com/awslabs/aws-crt-builder/blob/main/setup.py#L17) is specified in the setup script. 

#### Other related things:

3. Rather than relying on `scripts` you might consider using `console_scripts` and renaming `__main__.py` to just `builder.py` (you can even move this to `bin/builder.py` if you want to make the distinction even more clear)
```py
    entry_points = {
        'console_scripts': [
            'builder = builder.builder:main'
        ]
    },

```
This would require a small structural change to the main script to introduce the `main` function and call it like so
```py
if __name__ == '__main__':
    main()
```
You can leave it as scripts if you want but this has the advantage of not needing to specify the shebang `#!/usr/bin/env python` in the script file (which is currently missing and causes issues).

4. There is no need to be messing with [sys.path](https://github.com/awslabs/aws-crt-builder/blob/main/builder/__main__.py#L11) in the main script. This is likely due to a structural issue with the package or how you are hacking on it locally. You can resolve this by creating a proper package and then using full imports everywhere:

e.g. instead of `from actions.script import Script` you would do `from builder.actions.script import Script`

To hack on it locally just do `pip install -e .` to install the entire package in "editable" mode. Your build script will see all the reflections to any changes made locally.

5. `from __future__ import print_function` is unnecessary in Python3

[...]

PR #91, opened 2021-01-21
refactor project structure

### LINKED ISSUE #89: import JDK8 fails if toolchain not set
Importing `JDK8` when `needs_compiler` is false throws an exception:

```
Traceback (most recent call last):
  File "/opt/python/cp39-cp39/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/python/cp39-cp39/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/bin/builder.pyz/__main__.py", line 261, in <module>
  File "/usr/local/bin/builder.pyz/__main__.py", line 96, in run_build
  File "/usr/local/bin/builder.pyz/__main__.py", line 51, in run_action
  File "/usr/local/bin/builder.pyz/scripts.py", line 151, in run_action
  File "/usr/local/bin/builder.pyz/actions/script.py", line 54, in run
  File "/usr/local/bin/builder.pyz/scripts.py", line 151, in run_action
  File "/usr/local/bin/builder.pyz/actions/script.py", line 56, in run
  File "/usr/local/bin/builder.pyz/__main__.py", line 70, in pre_build
  File "/usr/local/bin/builder.pyz/project.py", line 402, in pre_build
  File "/usr/local/bin/builder.pyz/project.py", line 224, in _build_project
  File "/usr/local/bin/builder.pyz/imports/jdk.py", line 45, in install
AttributeError: 'Env' object has no attribute 'toolchain'
Finished: DownloadDependencies
```

builder.json:

```json
{
    "needs_compiler": false,
    "imports": ["JDK8"]
}
```

Fails [here](https://github.com/awslabs/aws-crt-builder/blob/main/builder/imports/jdk.py#L45) because `env.toolchain` [isn't set](https://github.com/awslabs/aws-crt-builder/blob/main/builder/__main__.py#L250) when `needs_compiler` is false.

aajtodd:
*Description of changes:*
* ~(refactor) Execute steps from the working directory of the project the step belongs to (as opposed to running all steps from the root project)~
* ~(feat) Add `pre_build_steps` and `post_build_steps` to `upstream` dependencies. These steps run before/after the upstream project is built.~
* (fix) `JDK8` import when toolchain is not set (i.e. `needs_compiler=false`). (fixes #89)
* (feat) Add `env` key to config that applies to all steps
* (refactor) Structural changes (some of these were brought up in #88)
    * Restructure project imports to work as a normal python package. Before there was a mix of paths used before (e.g. `import core.action` vs `import builder.core.action` as well as `sys.path` manipulation. This can cause all sorts of problems (e.g. `instanceof` checks can fail as it will see a module/symbol as `xyz.foo` instead of `mypackage.xyz.foo`). 
    * Rename cli entrypoint from `__main__.py` to just `main.py`; zipapp allows you to specify a function as an entrypoint and will create the `__main__.py` script for you.
    * Add console script to `setup.py`. Now when you install locally (`pip install .` or `pip install -e .` for development) to your virtualenv it will add `builder` to your path so you can use it as a normal exe.
* (fix) branch checkout swallowing errors and printing out the wrong branch name when it doesn't exist
* (fix) CLI help flag/usage `-h` not working
* (fix) various usages of `env.config` vs `self.config`
    * Fixes `downstream` consumer tests not running
* (fix) determining branch name when executed from Github `pull_request` event/workflow
* (test) Added some basic unit tests we can hopefully build on
* (fix) builder arguments duplicated in `entrypoint.sh` when `--version` is set
* (feat) define `{project_source_dir}` variable which will always be the project being built (`{source_dir}` and `{project_source_dir}` are the same for the "root" project but downstream/upstream projects `{source_dir}` currently refers to the root project `{source_dir}`).

### DIFF
--- a/.github/workflows/create-channel.yml
+++ b/.github/workflows/create-channel.yml
@@ -44,7 +44,9 @@ jobs:
       run: |
         export CHANNEL=${{ steps.tag.outputs.release_tag }}
         mkdir -p build
-        python3 -m zipapp --python="/usr/bin/env python3" --output=build/builder builder
+        mkdir -p staging
+        cp -r builder staging/.
+        python3 -m zipapp --python="/usr/bin/env python3" -m "builder.main:main" --output=build/builder staging
         aws s3 cp build/builder s3://$AWS_S3_BUCKET/channels/$CHANNEL/builder.pyz
 
     - name: Artifact builder
--- a/builder/core/env.py
+++ b/builder/core/env.py
@@ -95,12 +95,12 @@ class Env(object):
         self.config = self.project.get_config(self.spec, self.args.cli_config)
 
         # Once initialized, switch to the source dir before running actions
-        self.source_dir = os.path.abspath(self.project.path)
-        self.variables['source_dir'] = self.source_dir
-        self.shell.cd(self.source_dir)
+        self.root_dir = os.path.abspath(self.project.path)
+        self.variables['root_dir'] = self.root_dir
+        self.shell.cd(self.root_dir)
 
         # Allow these to be overridden by the project, and relative to source_dir if not absolute paths
-        build_dir = self.config.get('build_dir', os.path.join(self.source_dir, 'build'))
+        build_dir = self.config.get('build_dir', os.path.join(self.root_dir, 'build'))
         self.build_dir = os.path.abspath(build_dir)
         self.variables['build_dir'] = self.build_dir
 
@@ -112,12 +112,12 @@ class Env(object):
         self.install_dir = os.path.abspath(install_dir)
         self.variables['install_dir'] = self.install_dir
 
-        print('Source directory: {}'.format(self.source_dir))
+        print('(root) Source directory: {}'.format(self.root_dir))
         print('Build directory: {}'.format(self.build_dir))
 
         Project.search_dirs += [
             self.build_dir,
-            self.source_dir,
+            self.root_dir,
             self.deps_dir,
         ]
 
--- a/builder/core/env.py
+++ b/builder/core/env.py
@@ -47,7 +47,7 @@ class Env(object):
             if prev:
                 if not isinstance(prev, list):
                     prev = [prev]
-                    val = prev.push(val)
+                    val = prev.append(val)
             self.config[key] = val
 
         # build environment set up
--- a/builder/core/env.py
+++ b/builder/core/env.py
@@ -146,13 +146,13 @@ class Env(object):
             # and we need to grab the branch being merged from
             # see: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
             branch = github_head_ref
-            print(f"Found github ref for PR from: {branch}")
+            print("Found github ref for PR from: {}".format(branch))
             return branch
         elif github_ref:
             origin_str = "refs/heads/"
             if github_ref.startswith(origin_str):
                 branch = github_ref[len(origin_str):]
-                print(f"Found github ref: {branch}")
+                print("Found github ref: {}".format(branch))
                 return branch
 
         try:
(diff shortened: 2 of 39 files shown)


builder/core/project.py
```diff
@@ -410,9 +427,32 @@ def pre_build(self, env):
         build_deps = []
         for d in deps:
             dep_steps = _build_project(d, env)
+
             if dep_steps:
                 build_deps += [Script(dep_steps,
                                       name='build {}'.format(d.name))]
+
+            # allow root project to register additional pre/post build steps that run before or after the dependency is built
```
justinboswell, line 435: So this will work, but I kind of wonder if we shouldn't just take the whole config tree and apply it, by ensuring that the upstream and downstream packages are cached with their contents before being resolved (which will cause a merge when the actual builder.json is loaded)

builder/imports/__init__.py
```diff
@@ -10,21 +10,21 @@
 import zipfile
 
 modules = []
-parent_package = None
+# parent_package = None
```
justinboswell, line 13: can these commented out lines just be removed?

.github/workflows/create-channel.yml
```diff
@@ -44,7 +44,9 @@ jobs:
       run: |
         export CHANNEL=${{ steps.tag.outputs.release_tag }}
         mkdir -p build
-        python3 -m zipapp --python="/usr/bin/env python3" --output=build/builder builder
+        mkdir -p staging
```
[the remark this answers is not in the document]
aajtodd: no on L49 we output to build/builder using staging as the directory to create the zipapp from.

zipapp takes the inner directory contents so if you do:

```
python3 -m zipapp --output=build/builder staging
```

It will create a zip archive of everything under `staging`. This is necessary for the modules to work as expected such that you end up with a ziparchive with the following structure:
```
builder/
    __init__.py
    main.py
    ...
```

before we were doing `python3 -m zipapp --output=build/builder builder` but this will leave out the top level `builder` directory/module which screws up the imports. 


We could probably just do this though instead which removes the need for us to do the copy manually or think about it:

```
pip install --target staging .
python3 -m zipapp --python="/usr/bin/env python3" -m "builder.main:main" --output=build/builder staging
```

Same thing without the manual copy and let's python figure out what the package is, but it has additional files compared to what we were doing so I tried to keep it similar. 

This is what you would do if we had dependencies though for example:
```
pip install -r requirements.txt --target staging
```

justinboswell (approved):

builder/core/env.py
```diff
@@ -112,12 +112,12 @@ def __setitem__(self, item, value):
         self.install_dir = os.path.abspath(install_dir)
         self.variables['install_dir'] = self.install_dir
 
-        print('Source directory: {}'.format(self.source_dir))
+        print('(root) Source directory: {}'.format(self.root_dir))
```
justinboswell, line 115: This can just be Root directory I think

tests/data/lib-1/builder.json
```diff
@@ -1,10 +1,15 @@
 {
     "name": "lib-1",
+    "variables": {
+        "gradlew": "{source_dir}/gradlew"
+    },
     "build_steps": [
-        "echo \"build lib-1\""
+        "echo \"build lib-1\"",
+        "{source_dir}/gradlew"
```
justinboswell, line 8: This should be `{gradlew}` right?

justinboswell (approved):
Fix-n-ship!

tests/data/lib-1/builder.json
aajtodd: I think this stemmed from two separate tests. I originally just had this and then realized it would be convenient to define a variable using `{source_dir}`. I can probably rip that first test out and remove this given that the other test now covers this I think.

PR #93, opened 2021-02-09
Added PHP 5.6 and PHP 7.4 docker images

justinboswell:

### DIFF
--- a/.github/docker-images/php5-linux-x64/Dockerfile
+++ b/.github/docker-images/php5-linux-x64/Dockerfile
@@ -20,6 +20,7 @@ RUN yum -y update \
     openssl-devel \
     libcurl-devel \
     libffi-devel \
+    libxml2-devel \
     && yum clean all \
     && rm -rf /var/cache/yum \
     && ln -s /usr/bin/cmake3 /usr/bin/cmake \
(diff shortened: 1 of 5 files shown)


.github/docker-images/php5-linux-x64/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+###############################################################################
+RUN python3 -m pip install --upgrade pip setuptools virtualenv \
+    && python3 -m pip install --upgrade awscli \
+    && aws --version
+
+###############################################################################
+# Install pre-built OpenSSL
+###############################################################################
+WORKDIR /tmp
+RUN curl -sSL https://d19elf31gohf1l.cloudfront.net/_binaries/libcrypto/libcrypto-1.1.1-linux-x64.tar.gz -o libcrypto.tar.gz \
```
sdavtaker, line 43: Sounds like pretty hardcoded that cloudfront address, hard to change in the future, maybe we can use a CNAME?

sdavtaker (approved):

.github/docker-images/php5-linux-x64/Dockerfile
justinboswell: We don't currently control a domain that we could put the CNAME in.

PR #96, opened 2021-02-13
Improved handling of cli config args.

graebm:
*Issue:*

wanted to pass `cmake_args=-DBUILD_SHARED_LIBS=ON` to builder and it didn't work right

*Description of changes:*

- move all logic for fixing up cli_config to one place
- consult default KEYS to see if it should be a list
- fix bugs when reading in list args
- assume user wants to "!" override a list. Allow "+" for appending to defaults.

### DIFF
--- a/builder/core/project.py
+++ b/builder/core/project.py
@@ -45,7 +45,6 @@ def _apply_value(obj, key, new_value, apply_before=True):
     if key_type == list:
         # apply the config's value before the existing list
         val = obj[key]
-        new_value = list(new_value)
         if apply_before:
             obj[key] = new_value + val
         else:
--- a/builder/core/project.py
+++ b/builder/core/project.py
@@ -45,14 +45,11 @@ def _apply_value(obj, key, new_value, apply_before=True):
     if key_type == list:
         # apply the config's value before the existing list
         val = obj[key]
-        if val:
-            new_value = list(new_value)
-            if apply_before:
-                obj[key] = new_value + val
-            else:
-                obj[key] = val + new_value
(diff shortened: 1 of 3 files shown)


builder/core/project.py
```diff
@@ -45,14 +45,11 @@ def _apply_value(obj, key, new_value, apply_before=True):
     if key_type == list:
         # apply the config's value before the existing list
         val = obj[key]
-        if val:
-            new_value = list(new_value)
-            if apply_before:
-                obj[key] = new_value + val
-            else:
-                obj[key] = val + new_value
+        new_value = list(new_value)
```
justinboswell, line 48: One issue with this vs the way it was in my PR is that list('test') creates ['t','e','s','t'] whereas the type that tests the rhs will always produce a list without restructuring strings

graebm: The diff is deceiving here. I just removed a branch from this function that could never be hit. I didn't change its logic at all

The `if val:` branch was always getting hit (we tested this earlier in the function) so I removed this check and the `else:` branch

graebm: strings getting ['s', 'p', 'l', 'i', 't'] is the bug that led me to make this PR

I address the issue in main.py (for CLI args) by consulting with the default keys, and making sure lists are treated as lists.

I honestly don't understand why this line `new_value = list(new_value)` is here at all

graebm: should the line be changed to this?
```
if not isinstance(new_value, list):
    new_value = [new_value]
```

justinboswell (approved):

PR #99, opened 2021-03-02
--cmake-extra lets you add extra cmake args to every project

graebm:
The existing `cmake_args` in the config does not affect the building of upstream dependencies

So add new CLI argument we can use like `--cmake-extra=-DBUILD_SHARED_LIBS=ON` and build everything in the tree with a given flag

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -142,8 +142,9 @@ def parse_args():
     parser.add_argument('--target', type=str, help="The target to cross-compile for (e.g. android-armv7, linux-x86, linux-aarch64)",
                         default='{}-{}'.format(current_os(), current_arch()),
                         choices=data.PLATFORMS.keys())
-    parser.add_argument('--cmake-extra', action='append', help='Extra cmake config arg applied to all projects.' +
-                        ' May be specified multiple times. (e.g "--cmake-extra=-DBUILD_SHARED_LIBS=ON"')
+    parser.add_argument('--cmake-extra', action='append', default=[],
+                        help='Extra cmake config arg applied to all projects. May be specified multiple times. ' +
+                        '(e.g "--cmake-extra=-DBUILD_SHARED_LIBS=ON"')
     parser.add_argument('args', nargs=argparse.REMAINDER)
 
     # hand parse command and spec from within the args given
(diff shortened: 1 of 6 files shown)


builder/main.py
```diff
@@ -142,6 +142,9 @@ def parse_args():
     parser.add_argument('--target', type=str, help="The target to cross-compile for (e.g. android-armv7, linux-x86, linux-aarch64)",
                         default='{}-{}'.format(current_os(), current_arch()),
                         choices=data.PLATFORMS.keys())
+    parser.add_argument('--cmake-extra', action='append', default=[],
```
justinboswell, line 145: This could just be parsed in cmake.py, and that way you don't have to deal with changes here.

See: https://github.com/awslabs/aws-crt-builder/blob/main/builder/actions/install.py#L35 for an example

builder/main.py
```diff
@@ -162,41 +165,40 @@ def parse_args():
             print('No command provided, should be [build|inspect|<action-name>]')
         sys.exit(1)
 
+    # parse the args we know, the rest end up in args.args for others to parse
+    args = parser.parse_args(argv)
```
justinboswell, line 169: This should be parse_known_args()

graebm: There are no unknown arguments because of this line (a few lines up):
`parser.add_argument('args', nargs=argparse.REMAINDER)`

that causes all unknown arguments to end up in `args.args`.

The old way did `args, extra = parser.parse_unknown_args()` but it was pointless, `extra` was always empty.

graebm: I guess I'm saying, we should use that argparse.REMAINDER thing, or use parse_unknown_args(), but doing both is pointless and confusing

justinboswell: oh I see, let's get rid of the remainder then, because actions depend on being able to find their args in there without modifying main

builder/main.py
graebm: Hmm.
I see your point modularity-wise
but that prevents the option from showing up in --help 😕
your call, but historically I've had a hard time knowing what the builder can or can't do

justinboswell: We could give actions the ability to add to the global arg parser, or allow them to fill in help info when necessary. I agree, it's a half-measure, but I think allowing actions to have their own args is pretty important, especially for use-cases like aws-crt-python where the entire build process is in an injected python script.

graebm: Moving it to cmake.py, adding to readme

justinboswell (approved):

PR #102, opened 2021-03-04
check for cmake3 alias

graebm:
also delete folder that was checked in by accident

### DIFF
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -19,9 +19,9 @@ def _find_cmake(sh):
         if cmake:
             # show version
             sh.exec(cmake, '--version', check=True)
-
             return cmake
     raise Exception("cmake not found")
+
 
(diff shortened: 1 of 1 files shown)


builder/actions/cmake.py
```diff
@@ -2,14 +2,27 @@
...
 import os
 from pathlib import Path
+import shutil
 
 from builder.core.action import Action
 from builder.core.toolchain import Toolchain
 from builder.core.util import UniqueList
 
 
+@lru_cache(1)
```
justinboswell, line 15: What is this MAGIC?

justinboswell (approved):

builder/actions/cmake.py
graebm: it caches the results of the function
so we only end up looking for cmake once

PR #106, opened 2021-03-11
Clang CI Temp Fix (hopefully)

bretambrose:
Temporary workaround for old clang fetch failures by switching from ubuntu 16 to 18 as the base image.  Yes the container is still called ubuntu-16-x64 despite this change because updating every single ci.yml sounds like misery for a temp change (until we get  docker images with clang included available)

### DIFF
--- a/.github/docker-images/ubuntu-16-x64/Dockerfile
+++ b/.github/docker-images/ubuntu-16-x64/Dockerfile
@@ -25,13 +25,14 @@ RUN apt-get update -qq \
 ###############################################################################
 # Python/AWS CLI
 ###############################################################################
-RUN python3 -m pip install setuptools
-RUN python3 -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


.github/docker-images/ubuntu-16-x64/Dockerfile
```diff
@@ -24,14 +25,17 @@ RUN apt-get update -qq \
 ###############################################################################
 # Python/AWS CLI
 ###############################################################################
-RUN python3 -m pip install setuptools \
-    && python3 -m pip install --upgrade awscli \
-    && aws --version
+RUN python3 -m pip install setuptools
```
justinboswell, line 28: Put these all in one command, otherwise docker will create a layer for each of these. That's why there's so many continuations in Dockerfiles

justinboswell (approved):

PR #118, opened 2021-05-11
Removing macOS11 from CI

DavidOgunsAWS:
*Description of changes:*
* Removing this step from crt-builder CI as we aren't ready for it yet


kneekey23:
actually if this using the virtual environments from github actions, we have been added to the private preview so you can actually leave this and not remove it if you wanted to :)

justinboswell:
We now have MacOS 11 access, see https://github.com/awslabs/aws-crt-builder/pull/128

PR #130, opened 2021-07-22
Install node on the manylinux aarch64 image

bretambrose:


TingDaoK (approved):
Looks good, just curious about why use this old version instead of some latest one?

bretambrose:
> Looks good, just curious about why use this old version instead of some latest one?

Well it's for CD so you want the old manylinux for oldest possible glibc compatibility.

PR #139, opened 2021-08-18
Linux flavor hosts

bretambrose:
* Host definitions for new linux flavors
* Disable warnings on s2n builds
* Update new linux flavor containers a little

### DIFF
--- a/builder/imports/s2n.py
+++ b/builder/imports/s2n.py
@@ -8,6 +8,10 @@ config = {
     'targets': ['linux', 'android'],
     'test_steps': [],
     'build_tests': False,
+    # s2n imports pq crypto code written by people who don't bother to match the function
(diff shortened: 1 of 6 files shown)


builder/imports/s2n.py
```diff
@@ -8,6 +8,7 @@
     'targets': ['linux', 'android'],
     'test_steps': [],
     'build_tests': False,
+    'cmake_args': ['-DUNSAFE_TREAT_WARNINGS_AS_ERRORS=OFF']
```
justinboswell, line 11: Could you add a comment as to why we are doing this here?

justinboswell (approved):
Fix-n-ship

builder/imports/s2n.py
bretambrose: Sure

PR #142, opened 2021-08-18
Upload simple ci-via-container script as part of release process

bretambrose:

### DIFF
--- a/.github/workflows/linux-container-ci.sh
+++ b/.github/workflows/linux-container-ci.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# $1 - Package Name
+# $2 - Builder version
+# $3 - Ecr Repo
+# $4 - Image Name
+
+set -e
+
+PACKAGE_NAME=$1
+shift
+BUILDER_VERSION=$1
+shift
+ECR_REPO=$1
+shift
+IMAGE_NAME=$1
+shift
+
+aws ecr get-login-password | docker login ${ECR_REPO} -u AWS --password-stdin
+export DOCKER_IMAGE=${ECR_REPO}/${IMAGE_NAME}:${BUILDER_VERSION}
+docker pull $DOCKER_IMAGE
+docker run --env GITHUB_REF $DOCKER_IMAGE --version=${BUILDER_VERSION} build -p ${PACKAGE_NAME} "$@"
--- a/.github/workflows/linux-container-ci.sh
+++ b/.github/workflows/linux-container-ci.sh
@@ -1,21 +1,14 @@
 #!/bin/bash
-# $1 - Package Name
-# $2 - Builder version
-# $3 - Ecr Repo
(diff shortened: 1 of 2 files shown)


.github/workflows/linux-container-ci.sh
```diff
@@ -0,0 +1,21 @@
...
+BUILDER_VERSION=$1
+shift
+ECR_REPO=$1
+shift
+IMAGE_NAME=$1
+shift
+
+aws ecr get-login-password | docker login ${ECR_REPO} -u AWS --password-stdin
+export DOCKER_IMAGE=${ECR_REPO}/${IMAGE_NAME}:${BUILDER_VERSION}
+docker pull $DOCKER_IMAGE
```
justinboswell, line 20: `docker run` will pull the image automatically.

.github/workflows/linux-container-ci.sh
```diff
@@ -0,0 +1,21 @@
...
+docker run --env GITHUB_REF $DOCKER_IMAGE --version=${BUILDER_VERSION} build -p ${PACKAGE_NAME} "$@"
```
justinboswell, line 21: I don't think you want `$@` quoted, that'll cause it to group things together in a weird way that may not parse well

.github/workflows/linux-container-ci.sh
```diff
@@ -0,0 +1,21 @@
...
+# $3 - Ecr Repo
+# $4 - Image Name
+
+set -e
+
+PACKAGE_NAME=$1
+shift
+BUILDER_VERSION=$1
+shift
+ECR_REPO=$1
```
justinboswell, line 13: Does the repo ever change? Couldn't we just embed this? It's not secret really, it requires auth.

.github/workflows/linux-container-ci.sh
justinboswell: Also, you could just have the args in the CI yaml start with build -p whatever whatever, so this line becomes
`docker run --env GITHUB_REF $DOCKER_IMAGE --version=${BUILDER_VERISON} $@`

.github/workflows/linux-container-ci.sh
bretambrose: Will update

.github/workflows/linux-container-ci.sh
bretambrose: Unsure about the quotes.  Everything I read said to use them (https://stackoverflow.com/questions/1537673/how-do-i-forward-parameters-to-other-command-in-bash-script) and a run of a multi-param invocation seems to pass them properly: https://github.com/awslabs/aws-c-io/runs/3365801050#step:2:616

bretambrose: Remove package_name as an arg

justinboswell (approved):

PR #150, opened 2021-11-02
Added support for Apple watchOS and tvOS

justinboswell:

### DIFF
--- a/builder/core/data.py
+++ b/builder/core/data.py
@@ -603,6 +603,12 @@ PLATFORMS = {
     'ios-armv8': {
         'cross_compile_platform': None
     },
+    'tvos-armv8': {
+        'cross_compile_platform': None
+    },
+    'watchos-armv8': {
+        'cross_compile_platform': None
+    }
     # Linux is done procedurally, below
 }
 
--- a/builder/core/data.py
+++ b/builder/core/data.py
@@ -335,7 +335,7 @@ TARGETS = {
         'variables': {
             'exe': '',
             'osx_deployment_target': '13.0',
-            'osx_architectures': 'arm64;x86_64'
+            'osx_architectures': 'arm64'
         },
     },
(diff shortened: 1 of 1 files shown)


builder/core/data.py
```diff
@@ -338,6 +338,48 @@ class PKG_TOOLS(Enum):
...
+        'architectures': {
+            'armv8': {
+                'variables': {
+                    'osx_architectures': 'arm64'
+                }
+            }
+        },
+        'variables': {
+            'exe': '',
+            'osx_deployment_target': '13.0',
```
kneekey23, line 358: dont think this is correct. the latest version of tvOS is 7.0 and we should proably target two before so 5.0. 13 is the iOS version I believe.

builder/core/data.py
```diff
@@ -338,6 +338,48 @@ class PKG_TOOLS(Enum):
...
```
kneekey23, line 379: same here. think this should be 5.0

builder/core/data.py
```diff
@@ -338,6 +338,48 @@ class PKG_TOOLS(Enum):
...
+            'osx_architectures': 'arm64;x86_64'
```
kneekey23, line 359: can the architecture ever be x86_64?

builder/core/data.py
```diff
@@ -338,6 +338,48 @@ class PKG_TOOLS(Enum):
...
```
kneekey23, line 380: same question here as above?

builder/core/data.py
justinboswell: good catch

kneekey23 (approved):
fix n ship and 2 questions

builder/core/data.py
justinboswell: tvOS tracks iOS, this one is still right :)

PR #153, opened 2021-11-04
Fix release/channel ECR repo usage

justinboswell:


kneekey23 (approved):
interesting how they all became slashes when they should be dashes but lets go with it

bretambrose (approved):

PR #168, opened 2022-03-22
Attempt to restore manylinux1 image build and publish

bretambrose:

### DIFF
--- a/.github/docker-images/manylinux1-x64/Dockerfile
+++ b/.github/docker-images/manylinux1-x64/Dockerfile
@@ -1,0 +1,35 @@
+# See: https://github.com/pypa/manylinux	FROM docker.pkg.github.com/awslabs/aws-crt-python/manylinux1:x64
+# and: https://github.com/pypa/python-manylinux-demo
+FROM quay.io/pypa/manylinux1_x86_64
+
+###############################################################################
+# Basics
+###############################################################################
+RUN yum -y update \
+    && yum -y install sudo \
+    && yum clean all
+
+###############################################################################
+# Python/AWS CLI
+###############################################################################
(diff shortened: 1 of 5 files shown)


.github/docker-images/manylinux1-x64/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# See: https://github.com/pypa/manylinux	FROM docker.pkg.github.com/awslabs/aws-crt-python/manylinux1:x64
+# and: https://github.com/pypa/python-manylinux-demo
+FROM quay.io/pypa/manylinux1_x86_64
```
TwistedTwigleg, line 3: Will manylinux1 being not actively supported be an issue? From what I can tell, it seems to mostly apply to using Python 3.10+, so my understanding is that so long as we do not need Python 3.10+ it should be good?

TwistedTwigleg: Did some additional reading and from what I can tell reading the issue on dropping mannylinux1 support on the MannyLinux repository, it looks like it is still getting automatic updates and builds. It seems that as long as we do not need the latest Python and Pip versions for this particular DockerFile, it should be fine.

TingDaoK: Yeah, it doesn't have python 3.10. We are not testing python 3.10, so, we are good. And, we will drop it as soon as CLI drops the support.

TingDaoK (approved):
🙆‍♂️

.github/docker-images/manylinux1-x64/Dockerfile
TwistedTwigleg: Ah okay, good to know - Thanks!

PR #171, opened 2022-04-28
Remove PHP docker images

graebm:
These aren't building due to latest Composer dropping support for older PHP versions.

Also, these don't seem to be used anywhere. aws-crt-php used to use them, but doesn't anymore (see https://github.com/awslabs/aws-crt-php/pull/41)

So I'm just deleting them.
    

    
Red code is best code.


TingDaoK (approved):
:shipit:

bretambrose (approved):

PR #198, opened 2022-10-24
Stop using python 3.6, which is long past end of life.

graebm:
**Issue:**

aws-crt-python is removing support for python3.6: https://github.com/awslabs/aws-crt-python/pull/402
but several of our docker images are still using python3.6

**Description of change:**

For distros where python3.6 was the default (ubuntu18, rhel8), switch to using the newest python version available to that distro.

### DIFF
--- a/.github/docker-images/rhel8-x64/Dockerfile
+++ b/.github/docker-images/rhel8-x64/Dockerfile
@@ -2,7 +2,17 @@ FROM registry.access.redhat.com/ubi8/ubi:latest
 
 SHELL ["/bin/bash", "-c"]
 
-RUN dnf install -y gcc gcc-c++ cmake wget python39 git make sudo python39-devel
+RUN dnf install -y \
+    gcc \
+    gcc-c++ \
+    cmake \
+    wget \
+    git \
+    make \
+    sudo \
+    # RHEL8's default python3 is python3.6, which is EOL.
+    # So install python3.9 (latest version on this distro, circa Oct 2022)
+    python39-devel
 
 ###############################################################################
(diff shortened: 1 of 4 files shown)


.github/docker-images/rhel8-x64/Dockerfile
```diff
@@ -2,7 +2,17 @@ FROM registry.access.redhat.com/ubi8/ubi:latest
...
+    gcc \
+    gcc-c++ \
+    cmake \
+    wget \
+    git \
+    make \
+    sudo \
+    # RHEL8's default python3 is python3.6, which is EOL.
+    # So install python3.9 (latest version on this distro, circa Oct 2022)
+    python39-devel
```
waahm7, line 15: I think this should have python3-pip

graebm: On this distro, you can install the one package `python39-devel` and it ends up installing the other python stuff like pip as a dependency. So I was just simplifying:
```
Installing:
 python39-devel              
Installing dependencies:
 libpkgconf                  
 pkgconf                     
 pkgconf-m4                  
 pkgconf-pkg-config          
 python39                    
 python39-libs               
 python39-pip-wheel          
 python39-setuptools-wheel   
Installing weak dependencies:
 python39-pip                
 python39-setuptools         
Enabling module streams:
 python39                    
```

In other distros you need to install them separately

If Linux variants all made a big effort to do everything the same ... you wouldn't have so many Linux variants

waahm7 (approved):

TingDaoK (approved):

PR #202, opened 2022-11-14
Pull Node x86 build for x86 platform

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Pull Node build from https://unofficial-builds.nodejs.org/ for x86 platforms

### DIFF
--- a/builder/imports/nodejs.py
+++ b/builder/imports/nodejs.py
@@ -50,10 +50,10 @@ class NodeJS(Import):
                 append_times = 0
                 while re.match('^([0-9]+\.){2}[0-9]+$', v) == None:
                     # Only try append sub version twice
-                    if append_times < 2 :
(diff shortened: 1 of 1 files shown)


builder/imports/nodejs.py
```diff
@@ -44,12 +47,15 @@ def install(self, env):
         self.install_dir = os.path.join(env.deps_dir, self.name)
         sh.mkdir(self.install_dir)
 
-        if current_os() == 'windows':
-            self.install_nvm_choco(env)
+        if current_arch() == "x86":
+            self.install_node_build(env)
```
graebm, line 51: trivial: maybe name this `install_node_via_unofficial_build()`

this is similar to the existing `install_node_via_nvm()` function

also just calling it "build" made me thing it was something we were compiling locally

PR #212, opened 2022-12-08
add --coverage

TingDaoK:
- Add `--gcc-coverage`, to generate code coverage repo if available. 
- Only works for using Cmake and GCC as compiler. 
- We could involve more time to make a more generic solution for different compiler, but for real, we only need one working case for the report to work. So, it's not really worth to put time on it for now. We can come back later if we need more case.

### DIFF
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -158,6 +158,10 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
     if coverage:
         if c_path and "gcc" in c_path:
             # Tell cmake to add coverage related configuration. And make sure GCC is used to compile the project.
+            # CMAKE_C_FLAGS for GCC to enable code coverage information.
+            # COVERAGE_EXTRA_FLAGS="*" is configuration for gcov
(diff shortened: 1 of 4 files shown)


TingDaoK:
Talked with Bret, aws-crt-nodejs downstream test is failing because of the AWS role builder using doesn't have permission to do MQTT5, we believe.

Bret will work on a fix, but for now, let's ignore it.

builder/actions/cmake.py
```diff
@@ -101,7 +101,7 @@ def _project_dirs(env, project):
     return source_dir, build_dir, install_dir
 
 
-def _build_project(env, project, cmake_extra, build_tests=False, args_transformer=None):
+def _build_project(env, project, cmake_extra, gcc_coverage=False, build_tests=False, args_transformer=None):
```
waahm7, line 104: I think we should always add new parameters at the end, so it doesn’t disturb legacy code.

TingDaoK: sure, but it's a private function, no one beside this package uses it.

sbSteveK (approved):

PR #213, opened 2022-12-13
add include and exclude config for coverage report

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder/main.py
+++ b/builder/main.py
@@ -17,6 +17,7 @@ from builder.core.project import Project
 from builder.core.scripts import Scripts
 from builder.core.toolchain import Toolchain
 from builder.core.host import current_os, current_host, current_arch, current_platform, normalize_target
+from builder.core.util import UniqueList
 import builder.core.data as data
 
 import builder.core.api  # force API to load and expose the virtual module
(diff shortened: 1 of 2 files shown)


builder/main.py
```diff
@@ -145,7 +146,14 @@ def parse_args():
     parser.add_argument('--variant', type=str, help="Build variant to use instead of default")
     parser.add_argument('--cmake-extra', action='append', default=[])
     parser.add_argument('--coverage', action='store_true',
-                        help="Enable test coverage report and upload it the codecov. Only supported when using cmake with gcc as compiler, error out on other cases.")
+                        help="Enable test coverage report and upload it the codecov. Only supported when using cmake with gcc as compiler, error out on other cases.\n"
+                        + "Use --coverage-inlucde and --coverage-exclude to report the needed coverage file. The default code coverage report will include everything in the `source/` directory")
```
graebm, line 150:
```suggestion
                        + "Use --coverage-include and --coverage-exclude to report the needed coverage file. The default code coverage report will include everything in the `source/` directory")
```

builder/main.py
```diff
@@ -234,8 +242,21 @@ def upload_test_coverage(env):
     # only works for linux for now
     env.shell.exec('curl', '-Os', 'https://uploader.codecov.io/latest/linux/codecov', check=True)
     env.shell.exec('chmod', '+x', 'codecov', check=True)
+    include_args = UniqueList(['source/'])
+    include_args += env.args.coverage_include
+    exclude_args = UniqueList(env.args.coverage_exclude)
+    uploader_args = []
+    for include in include_args:
+        include += '*'
```
graebm, line 250: trivial: this seems to assume that the user passed a path ending with "/" ?

maybe don't add '/' or '*' and simply document the path should be like "source/*"

that also lets people include or exclude specific files, instead of assuming it's always a folder

graebm (approved):
fix & ship

builder/main.py
graebm: actually ... there may be an issue asking users to pass "*" since it needs to be quoted to avoid the shell doing magic on that character

python's core libraries *should* be good about adding those quotes, but who knows what kind of assumptions our code makes passing around args...

test this branch against aws-c-common (using BUILDER_SOURCE=channels) before doing an actual release with my suggestions

and if it doesn't work, go ahead and keep adding the "*" automatically

TingDaoK: actually when people want to include/exclude a specific file, the asterisk is also kind needed as the file to upload ends with `.gcov`. (I was checking the `.endwith("/")` to add `*` in the end, and then find out that the files cannot be found).

PR #214, opened 2023-01-11
don't repeat output of failed command

graebm:
**Issue:**
The output of a failing command prints twice, making the logs hard to parse.

**Diagnosis:**
It prints once when the command runs, then the output is added to an exception which gets printed out again.

I don't know why the output was ever added to the exception. This was changed 1.5 years ago, but there's no real explanation why in the PR: https://github.com/awslabs/aws-crt-builder/pull/129

This also risks leaking secrets when the command is run with `quiet=True`

**Description of changes:**
- Remove output from Exception
- Use named keyword arguments instead of `**kwargs` so it's easier to figure out what all the options are
- Helpers for stringifying commands args

### DIFF
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -39,7 +39,7 @@ def cmake_path(cross_compile=False):
 def cmake_version(cross_compile=False):
     if cross_compile:
         return '3.17.1'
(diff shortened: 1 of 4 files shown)


TingDaoK (approved):
I SAW DOUBLE LOG ALL THE TIME AND THAT'S IT!
AWESOME!!!!!!! LOVE IT!

PR #191, opened 2022-10-04
Added Raspberry Docker

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Added the docker image for Raspberry Bullseye. Both images are built from Raspbian Stretch based on https://github.com/navikey/raspbian-bullseye

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #111, opened 2021-04-06
Added armv5 and armv7 debian stretch images (closest to raspberry pi OS)

justinboswell:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #220, opened 2023-03-09
Unbreak builder on Python < 3.6

knightjoel:
*Issue #, if available:*

N/A

*Description of changes:*

Python f-strings became available in Python 3.6. Some of the CI jobs (eg, linux-compat/debian-stretch-arm32* in c-common) use a version of Python < 3.6. This PR allows builder to work in those jobs.

### DIFF
--- a/builder/actions/script.py
+++ b/builder/actions/script.py
@@ -56,7 +56,7 @@ class Script(Action):
             elif callable(cmd):
                 children += to_list(cmd(env))
             else:
-                print('Unknown script sub command: {}: {}', cmd_type, cmd)
+                print('Unknown script sub command: {}: {}'.format(cmd_type, cmd))
                 sys.exit(4)
         return children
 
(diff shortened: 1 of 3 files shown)


graebm:
Python 3.6 is end-of-life
Debian Stretch is end-of-life

I strongly prefer to NOT support things beyond their end of life. It can be very unpleasant (can't use newer features, spend a lot of time trying to keep it working as things you depend on drop support and stop working)

I'd prefer we remove these docker images altogether. Looking at the PR where they were added 2 years ago, it seems like we were trying to get coverage for Raspberry Pi, and this was the best we could do.
https://github.com/awslabs/aws-crt-builder/pull/111
> Added armv5 and armv7 debian stretch images (closest to raspberry pi OS)

But a few months back we added a better `raspbian-bullseye` docker image
https://github.com/awslabs/aws-crt-builder/pull/191

Let's replace uses of the old images with the new image

knightjoel:
That makes perfect sense. I'll leave builder alone for now, until I know nothing is left that uses these older images.

PR #224, opened 2023-04-04
Fix AL2 docker image build

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/docker-images/al2-x64/Dockerfile
+++ b/.github/docker-images/al2-x64/Dockerfile
@@ -20,8 +20,6 @@ RUN yum -y update \
     which \
     && yum clean all \
     && rm -rf /var/cache/yum \
-    && ln -s /usr/bin/cmake3 /usr/bin/cmake \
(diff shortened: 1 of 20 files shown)


.github/docker-images/al2-x64/Dockerfile
```diff
@@ -21,15 +20,13 @@ RUN yum -y update \
     which \
     && yum clean all \
     && rm -rf /var/cache/yum \
-    && ln -s /usr/bin/cmake3 /usr/bin/cmake \
-    && ln -s /usr/bin/ctest3 /usr/bin/ctest \
```
waahm7, line 25: trivial: Why are we removing this? I don't see any error regarding this in build logs.

waahm7 (approved):

.github/docker-images/al2-x64/Dockerfile
DmitriyMusatkin: al2 cmake install already creates soft link, so it errors out when you try to create them again. there should be an old generation run where this error occured

PR #226, opened 2023-04-07
Action to Setup CI from Environment files

TwistedTwigleg:
*Description of changes:*

Adds a builder action called `SetupCIFromFiles` that will run formatted JSON files and add them as environment variables in a batch configuration. This is done to simplify setting up environment variables for testing across repositories, part of a project I am working on. The goal is to have ALL unit test input be environment variables, if at all possible, and use this builder action to make it as easy as possible to maintain.

This new action uses a new `builder.json` property called `CI_ENVIRONMENT_VARIABLE_FILES` that is an array containing all of the paths for the files and/or S3 URLs pointing to the JSON files based on the extension in the bucket. If there is no extension, then it will error out.

The builder action allows setting environment variables entirely from a JSON value, where the values for the environment variables can come from raw input (I.E in the JSON file), secrets, or download S3 files.

### Example builder config

~~~sh
{
    "name": "aws-crt-java",
    "CI_ENVIRONMENT_VARIABLE_FILES": [
       "file_path_here.json",
       "s3://url_here.json",
    ],
}
~~~

### Example JSON

**Note**: Comments in the JSON parser for this script have to be arrays of strings to be properly ignored, and these have to be children of the array. Note that prepending it with `[COMMENT]` is not required. The example below has comments showing how this is done.

~~~json
[
    ["[COMMENT] <input_data>: Will set environment variable <name> to whatever is in <input_data>"],
    {
        "name": "ENVIRONMENT_NAME_ONE",
        "input_data": true
    },
    ["[COMMENT] <input_secret>: Will set environment variable <name> to whatever data is in the",
     "secret with the name in <input_secret>"],
    {
        "name": "ENVIRONMENT_NAME_TWO",
        "input_secret": "secret_name_here"
    },
    ["[COMMENT] <input_s3>: Will download a S3 file at the given S3 URL to a temporary file and then will",
     "set environment variable <name> to the filepath to the temporary file"]
    {
        "name": "ENVIRONMENT_NAME_THREE",
        "input_s3": "s3://S3_URL_Here"
    },
    ["[COMMENT] If <file_tmp> is defined, it will set whatever data is in input",
     "(input_data, input_secret, etc) to",
     "a temporary file and set <name> to the filepath to the temporary file.",
     "Note: This is not needed for input_s3, which does this automatically"]
    {
        "name": "ENVIRONMENT_NAME_FOUR",
        "input_secret": "secret_name_here",
        "file_tmp": true
    },
   ["[COMMENT] If <input_role_arn> is defined, it will use the CLI to get the IAM",
    "role credentials for the given IAM role.",
    "The data will be in environment variables with the following names:",
    "    * <name>_ACCESS_KEY",
    "    * <name>_SECRET_ACCESS_KEY",
    "    * <name>_SESSION_TOKEN",
    "while <name> will be assigned to 'SUCCESS'"]
    {
        "name": "ENVIRONMENT_NAME_FIVE",
        "input_role_arn": "role arn here",
    },
    ["[COMMENT] <input_role_arn_secret> is the same as <input_role_arn>",
     "but it takes a AWS Secret Name instead of the raw IAM Role ARN"]
    {
        "name": "ENVIRONMENT_NAME_SIX",
        "input_role_arn_secret": "secret with role arn here",
    },
    ["[COMMENT] <os_only>: Checks the operating system and the desired operating",
     "system(s) and only sets the the environment variable if the current OS",
     "matches the desired.",
     "Valid values: <windows>, <mac>, or <linux>.",
     "Can include multiple via <,>. For example: <windows,linux>"],
    {
        "name": "ENVIRONMENT_NAME_SEVEN",
        "input_data": "example",
        "os_only": "windows" 
    },
    ["[COMMENT] <os_arm_skip>: Checks if the platform is ARM.",
     "If it is, then it does not set the environment variable."],
    {
        "name": "ENVIRONMENT_NAME_EIGHT",
        "input_data": "example",
        "os_arm_skip": "value does not matter"
    },
    ["[COMMENT] <os_codebuild_only>: Checks if the platform is Codebuild.",
     "If it is NOT Codebuild, then it does not set the environment variable."],
    {
        "name": "ENVIRONMENT_NAME_NINE",
        "input_data": "example",
        "os_codebuild_only": "value does not matter"
    },
]

~~~

_________


builder/core/data.py
```diff
@@ -29,6 +29,8 @@ class PKG_TOOLS(Enum):
     'cxx': None,  # c++ compiler
     'cmake_args': [],  # additional cmake arguments
 
+    'CI_ENVIRONMENT_VARIABLE_FILES': [],  # CI Environment Files files
```
DmitriyMusatkin, line 32: nit: with every other option being lowercase, this name looks really jarring

builder/actions/setup_ci_from_files.py
```diff
@@ -0,0 +1,301 @@
...
+import tempfile
+import xml.etree.cElementTree as xml
+
+
+class SetupCIFromFiles(Action):
+    current_folder = None
+    env_instance = None
+    tmp_file_storage = []  # NOTE: This is needed to keep the tmp files alive
+
+    def _is_same_os_as_string(self, os_string):
```
DmitriyMusatkin, line 18: most of the logic here seems to be the same as current_os helper we already have in host.py

builder/actions/setup_ci_from_files.py
```diff
@@ -0,0 +1,301 @@
...
+                if ('input_s3' in item):
+                    try:
+                        tmp_file = tempfile.NamedTemporaryFile()
+                        tmp_file.flush()
+                        self.tmp_file_storage.append(tmp_file)
+                        tmp_s3_filepath = tmp_file.name
+                        self.copy_s3_file(str(item['input_s3']), tmp_s3_filepath)
+                        environment_value = str(tmp_s3_filepath)
+                    except Exception as ex:
+                        sys.exit(f"[FAIL] {environment_name} [Input S3]: Exception ocurred trying to get S3 file")
```
DmitriyMusatkin, line 123: are users expected to have aws cli configured to be able to access account that owns the bucket? 
what if the user does not have those creds? 
who is putting files in s3 and how we are versioning them? I would make the argument that any configs outside of secrets should be in github

builder/actions/setup_ci_from_files.py
```diff
@@ -0,0 +1,301 @@
...
+
+                ############################################################
+                # OS Condition (starts with "os_")
+                ############################################################
+
+                # Checks to see if the operating system is the same as the desired operating system(s) of the environment variable.
+                # Valid JSON:
+                #   'os_only': "windows" or "linux" or "mac" <add ',' for multiple. Example: "windows,linux">
+                if ('os_only' in item):
```
DmitriyMusatkin, line 132: we already have concept of host/target configs in builder, why are we introducing another similar concept inside env config files?

builder/actions/setup_ci_from_files.py
TwistedTwigleg: Users are expected to have AWS CLI Configured with access to the bucket. If the account does not have access, then it will fail the builder action. This is the same for if the environment variable JSON/XML file is an S3 url: it assumes the user has access and will fail if there is an exception due to a lack of access.

For who is putting the files into S3: the idea is that our teams would do that. That way we can share the same configuration file across all 4 CRTs and 4 IoT SDKs when pointing to a common set of configuration options. For example, things like IoT certificates and private keys for unit-testing. Currently we share these files across the CRTs and SDKs, but how we get these files into the builder for unit testing vary completely by the CRT/SDK with no consistency. The idea is that we would have a single JSON/XML file that all the unit tests use across all 4 SDKs.

The reason for putting it in S3 instead of Secrets or GitHub is that way we can share it across repositories easier. We could put them in Secrets I suppose instead of S3 if that would be preferred.

______

Another reason we have `input_s3` is that some unit tests needs files that we cannot copy-paste into GitHub secrets or AWS Secrets due to their encoding. An example of this is PKCS12 keys and Java keystore files, as both have encoding that doesn't translate to straight text that can be inputted into AWS Secrets or GitHub secrets, so we have to download these files from S3.

builder/actions/setup_ci_from_files.py
TwistedTwigleg: The idea is that we use the exact same file for all of the CRTs and SDKs and for all platforms, therefore we need a way to differentiate between them when setting environment variables that are OS specific (like Windows certificate store, for example). I was initially going to do something like `+config` and `!config` so we could just add/remove/override the JSON/XML files, but as discussed in the internal thread on Slack, this would be complicated due to how the override system works. I thought it was just better to add the config check directly to the environment variables file.

builder/actions/setup_ci_from_files.py
DmitriyMusatkin: whats the proposed way people without access to our s3 are going to run builder? are we making those files public?

DmitriyMusatkin: a couple general thoughts. agree with the general idea for making tests more uniform, but i think we need to be really careful to avoid making it impossibly hard to run outside of ci
1. how are those resources in s3 created and maintained? relying on people manually uploading the files can be really flaky and not future proof (we've ran into similar issues with c-s3 before). I'm not against the idea in general, but i think any files uploaded to s3 need to be versioned and automatically generated as part of some release process (ex. be created on builder version release)
2.  can people run those tests without builder? outside of ci? builder should be a tool to make tests easier to run, but we should not make tests so complicated that its not feasible to run them without builder

builder/actions/setup_ci_from_files.py
DmitriyMusatkin: we support adding custom stuff pretest (initializing env variables, etc...), are there limitations with that, that make it not usable for env variable override? 
my concern with adding a whole new format for specifying those env variables is that its adding a lot of complexity to an already overly complex way to build thing

TwistedTwigleg: We could, but then we'd need a special builder action for each of the CI tests that have OS specific requirements. Granted sometimes this cannot be avoided, like with PKCS11 or Windows Cert Connect due to needing additional setup.

But on a whole, I think the idea is to have a single file that is the main set of environment variables so adding/removing/modifying environment variables is a one-stop-shop solution. Moving the specific environment variable setups into a bunch of builder actions would spread the setup of the environment variables across a bunch of different files.

That is not to say adding a way to change how the environment variable based on OS in the JSON/XML is the best solution, but I couldn't find a decent solution that still keeps the environment files all together and not duplicating builder action scripts across CRT and SDK repos.

I don't know... Will need to think further on this...

builder/actions/setup_ci_from_files.py
TwistedTwigleg: 1. The resources are just dumped into CI, I didn't add versioning or anything like that. You do bring up a good point about versioning and otherwise making it part of the release pipeline, as otherwise it would be too easy to change something and accidentally break multiple CI process across multiple repos.

2. Yeah, the tests can be run outside of the builder so long as the environment variables are set. What we could do to help people running it locally is make a "local" JSON/XML file that has example variables that they can fill-in/replace to allow the tests to run locally. While I have not made such file yet (I'm still trying to get it to run successfully and fully in the Java CRT repo on a custom branch) this is something I can look at doing.

_________

I agree that we need to be careful with this and thank you for bringing up these points. I think it will help future proof the solution moving forward and it highlights a few areas that still need thought/development.

builder/actions/setup_ci_from_files.py
```diff
@@ -0,0 +1,313 @@
...
+# SPDX-License-Identifier: Apache-2.0.
+
+from builder.core.action import Action
+from builder.core.host import current_os, current_arch
+import os
+import pathlib
+import sys
+import json
+import tempfile
+import xml.etree.cElementTree as xml
```
graebm, line 11: keep it simple
don't add complexity just because you can
we've never used XML for anything before, so don't start using it now

you can add comments in json like this:
`"_comment": "haha no one expected this field so I can put whatever I want here"`

requirements.txt
```diff
@@ -0,0 +1 @@
+PyYAML==6.0
```
graebm, line 1: builder had no external dependencies before and that was really nice

I don't see where this code is even using yaml?

builder/actions/setup_ci_from_files.py
```diff
@@ -0,0 +1,313 @@
...
+                        sys.exit(f"[FAIL] {environment_name} [Input S3]: Exception ocurred trying to get S3 file")
+
+                ############################################################
+                # OS Condition (starts with "os_")
+                ############################################################
+
+                # Checks to see if the operating system is the same as the desired operating system(s) of the environment variable.
+                # Valid JSON:
+                #   'os_only': "windows" or "linux" or "mac" <add ',' for multiple. Example: "windows,linux">
+                if ('os_only' in item):
```
graebm, line 129: After years of doing conditional stuff in builder.json files, trying to "program with data", my personal takeaway is: don't.

Don't take lessons from builder.json

This crazy system of trying to merge data, and mixing in conditional logic, is bad.

Just write a shell script or python script, where the conditional logic is in normal if-else statements. Scripts are easy to write, and they're easy for someone to understand and modify.

builder/actions/setup_ci_from_files.py
TwistedTwigleg: Will do. I will need to adjust the code a bit to skip fields starting with "_comment", but it is doable. I'll do such and remove XML parsing.

requirements.txt
TwistedTwigleg: This is an artifact back when it had YAML support. Will remove!

builder/actions/setup_ci_from_files.py
TwistedTwigleg: I get what you are saying, and maybe I'm misunderstanding, but:

1 | Shell scripts only run on Windows and the builder runs on all of the major operating systems.

I want to make whatever system we have for setting environment variables work across all builder supported operating systems. I was going to use a bunch of Shell scripts and just do it that way, but it won't work because of the lack of Windows support. Additionally, with how the builder works, we'd also need to somehow run `source Shell_Script_Here.sh` from the Python builder process AND have that set the environment variables in the builder, which I do not think will work...?

(Admittedly I still need to find a solution for Codebuild, where we currently don't seem to use the builder... I haven't quite figured this one out yet...)

2 |  Adding more scripts (Shell or Python) spreads the environment variables across multiple files rather than having one big file with everything included. The idea with this work is that it will allow us to have a single set of files that we can share across all CRTs and SDKs, rather than what we have now, where each one has a different setup and a different set of files.

I get avoiding complexity, but from a maintaince perspective I think we take a pretty big (and unwieldy) development cost if we add custom Python builder action scripts for each bit of setup. Sometimes it cannot be avoided, like with PKCS11 or Windows Certificate Store, as they need custom processing beyond just setting an environment variable.

However, for things like PKCS12 (MacOS only) and Custom Key Ops (Java and Linux only), both of these cases would need additional builder files. For Custom Key Ops it's arguably not a big deal since it's only on the Java CRT, but for PKCS12, this custom builder action/file would then need to be duplicated across each of the CRTs - so there would be 4 copies that essentially check if the OS is MacOS, download a file, and set a couple environment variables. Then if we need to change where the PKCS12 builder script to set a different environment variable name, for example, we'd need to make the change across the 4 CRTs.

This same issue with duplicate scripts would happen for the SDKs as well.

On a whole the idea with this is to try and consolidate testing setup into a single file (or small set of files for things that need additional processing, like PKCS11) that can be shared across all CRTs and SDKs. The idea is that we just point the builder action to the file(s) in S3, and then it sets up all the environment variables in the exact same way across the CRTs and SDKs.

To quote what I wrote earlier on this topic:

> We could, but then we'd need a special builder action for each of the CI tests that have OS specific requirements. Granted sometimes this cannot be avoided, like with PKCS11 or Windows Cert Connect due to needing additional setup.
>
> But on a whole, I think the idea is to have a single file that is the main set of environment variables so adding/removing/modifying environment variables is a one-stop-shop solution. Moving the specific environment variable setups into a bunch of builder actions would spread the setup of the environment variables across a bunch of different files.
>
> That is not to say adding a way to change how the environment variable based on OS in the JSON/XML is the best solution, but I couldn't find a decent solution that still keeps the environment files all together and not duplicating builder action scripts across CRT and SDK repos.

I understand trying to avoid complexity, but I don't think that adding a bunch of files that will be duplicated across the repositories is going to be easy to maintain.

3 | The system isn't merging data, it is just conditionally NOT setting data based on the OS, whether it's ARM or not, and/or whether the platform is in Codebuild.

I understand it does have conditional logic based on the OS, but I'm not sure it is more complex than having a bunch of builder scripts that check OS and conditionally set the environment variables... Likewise, I'm not sure they are harder to understand than something like `os_only: "windows` on a field in the JSON file.

___________

I'm not against the general idea of trying to make this as least complex as possible, but I'm not sure how to handle the conditional environment variables for OS without either making a BUNCH of scripts that are not as easily portable and reusable across CRTs and SDKs, or doing something like the conditional checks in the environment variables itself.

I might be missing something or misunderstanding what you are saying though.

graebm: I'm saying: trying to do conditional logic in a data file seems like a good idea at first, but quickly falls apart

Take a look at [python-v2's builder.json](https://github.com/aws/aws-iot-device-sdk-python-v2/blob/5bdbcc90256927a7d37cc9fc2b3adbad4e126dde/builder.json) and search the word "sucks". I desired to run a specific python installation, but builder.json didn't account for that. Look at all the hoops I had to jump through. Years later I solved the same problem in aws-crt-python by just moving all the logic [out of builder.json](https://github.com/awslabs/aws-crt-python/blob/80a1f21474096f6b54698a483ffe9b024ee43a3c/builder.json#L18-L21) and into a custom [python script](https://github.com/awslabs/aws-crt-python/blob/80a1f21474096f6b54698a483ffe9b024ee43a3c/.builder/actions/aws_crt_python.py#L229) where I could just do whatever craziness I needed in an obvious way. Zero regrets.

I'll do another post with my suggestion....

graebm: So the problem that aws-crt-builder solves is: it has common logic we re-use across repos, and it can run anywhere. Github actions are another way to re-use CI logic across repos, but Github actions can't run on Codebuild or on your personal machine.

The issue I have with aws-crt-builder is that it does a bunch of crazy data merging from builder.json that doesn't always work like you expect and simple logic bugs can take days to debug.

The problem you're having is: you have a bunch of logic you want to share across repos, but it only really applies to IoT repos, not EVERY repo that uses aws-crt-builder.

Maybe a somewhat dirty, but at least it's obvious, way of doing things is to add a file to aws-crt-builder that sets the environment variables they way you need for IoT. Something like: `aws-crt-builder/actions/iot_env_vars.py`. And every IoT repo can just run that action.

TwistedTwigleg: I think there might be a bit of a misunderstanding.

The plan is **NOT** to have every repository the builder uses run this new action. It is only for the CRT repositories and the IoT SDK repositories, as it's only for helping set up our IoT related tests ultimately. An example of the general idea, though its not in its final form yet, can be seen here on [this branch](https://github.com/awslabs/aws-crt-java/tree/unit_test_refactor).

(*In theory it could be used for the aws-c-* repositories too, but adjusting those tests and such to use the builder action is outside of the scope of the work assigned*).

The builder action will **not** invoke itself by default and run if `ci_environment_variable_files` is present, you still will have to call `Builder.setup_ci_from_files()`, like in [this code](https://github.com/awslabs/aws-crt-java/blob/unit_test_refactor/.builder/actions/aws_crt_java_test.py#L60).

Likewise, the files it processes are stored in the `builder.json` of each project ([example](https://github.com/awslabs/aws-crt-java/blob/unit_test_refactor/builder.json#L24-L26)). The data in the `builder.json` file is only for setting which file(s) to use for processing and I gave up on the idea of using something like `"+ci_environment_variable_files"` and `"!ci_environment_variable_files"` after seeing how complicate it was to implement such support into `builder.json`. The new action is entirely self contained in how it processes and deals with operating systems, it doesn't rely on the builder for anything more than checking the OS in `host.py`.

While I want to make this solution as generic as possible so EVERY repository **could** use it, I certainly do not want to force every builder repository to use it and therefore have it as opt-in only and configurable so if it is not needed it doesn't get in the way. It is NOT going to do any processing beyond setting environment variables and it will not install anything, it just pulls data and sets environment variables. For stuff that needs additional processing, like [Windows Certificate Store tests for example](https://github.com/awslabs/aws-crt-java/blob/unit_test_refactor/.builder/actions/windows_cert_test_setup.py), a separate builder action can (and should) be used for the reasons you mentioned: It is just easier and better to do complicated things that way.

I will write a document and share it internally with more information, a better explanation of how I plan the builder action to be used, and some examples of the data that could be passed (in addition to how customers might be able to use it and how we can share a 'dummy' configuration file). I would share it today, I just haven't had the time to sit down and write the file yet 😓

graebm: pretty much

I mean, that's essentially the same thing you wanted to accomplish, right? Have 1 file stored in S3, used by all the repos that run these MQTT integration tests?

graebm: I mean: the end goal is 1 file that specifies a bunch of custom stuff, that's used by several repos

but instead of a json file stored in S3, and a builder script that does custom processing on it

it's just a builder script

TwistedTwigleg: The general idea is the same, I agree. The thing I worry about with having it as a builder script is the maintenance cost, it being difficult to adjust quickly, and it would definitely make it harder for customers/others to run the builder action/tests.

________

For example, if we have it all in just a builder script, then making changes to environment variables requires the following:

Phase 1:
* (For the sake of this example, I'm treating all environment variables as key-value pairs for simplicity)
* Is the change on the key or value of the environment variable?
  * If the modification is for the key, you have to go through phase 2 below
  * If the modification is for a value, if the value is a secret or S3, you can just change the value in Secrets or S3. However, if the value is hard-coded, then you have to go to phase 2.
* Is the change on the additional processing of the environment variable? (PKCS11, Windows Key Cert, etc)
  * Have to modify the builder action and go through phase 2 below.

Phase 2:
* Make a PR with the change to the builder action and merge.
* Make a release of the builder with the newly edited action
* Go the the CRTs and/or SDKs that need the change and bump the version of the builder they use to the just released version
* Change complete!

While for the JSON file route, it instead looks like:
* (For the sake of this example, I'm treating all environment variables as key-value pairs for simplicity)
* Is the change on the key or value of the environment variable?
  * If so, change the JSON file accordingly, either in new JSON file or editing existing JSON file.
  * Change complete!
* Is the change on the additional processing of the environment variable? (PKCS11, Windows Key Cert, etc)
  * If so, there *should* be a builder action in the repo that does this additional processing, which may or may not be environment variable dependent. Change the builder action in the specific repo (or repos) accordingly.
  * Make a PR with change for builder action. Add/Change environment variables in JSON file if needed.
  * Change complete!

As you can see, it is a LOT easier to quickly make changes with the JSON version than the builder version. This can be seen as a con, but the thing I worry about with hard-coding the processing into the builder is that it means any time we have to make a small change, we have to release an entirely new builder version. This is slow and especially concerning if we ever had a security issue.

With a file based solution, especially if it's in S3, we can change the environment variables without needing to cut new releases, without needing to make 4+ PRs, and can adapt quickly. Again though, this could be seen as a con, as it also means we could break 4+ repositories without PRs, releases, etc. There is a trade off, and admittedly I'm not sure which is the best, but I think hard-coding or semi-hard-coding is going to make it hard to adjust.

Another thing I just thought about: It also means that before any new functionality can be tested in CI in the CRTs/SDKs that use environment variables, a PR has to be made on the builder that adds the needed material, a release of the builder made, and then the test can be run in CI on the CRT/SDK using the environment variables...

Overall, this just feels really slow and bulky to maintain, at least to me...

____________

Additionally, using a hard-coded builder action in the `aws-crt-builder` repository means for sure that customers cannot use the builder to setup and run the tests on their account, not without setting up every secret, S3 file, etc. With the JSON based solution, we could provide them a template JSON file they can just populate the values with to point to their resources, but with a hard-coded build action, I do not see how we could provide this as we'd need the customer to make their own fork of the builder, modify the action to their needs, and then use their custom fork of the builder instead of ours for any testing.

It also means that we have to use the same AWS account for the environment variables on the CRT and SDK repositories, not using the CRT account for CRT tests and the SDK account for SDK tests, as the builder will be either pointing to one or the other.
* Or in theory it could point to both or have a separate action for each, but again, this makes it harder to maintain and harder for customers.

__________

That is NOT to say what I have is perfect and I'm sure there is a middle ground we can reach, I just do not think that hard-coding it into a builder action is going to be a good idea for us to maintain nor for customers since it will make it impossible for them to run.

I think we need to find something that is:
* Easy to maintain, or at least minimally painful.
* Consolidated as much as possible - this is the ask/task that I am ultimately trying to perform.
* Opt-in to prevent users who are using the builder for non MQTT/IoT to still function
* Ideally allow the customer to be able to run our tests - I.E not find solutions that prevent the customer from running it.
  * We want to allow customers to run the test outside of the builder (which they can do via setting environment variables) and outside of CI if at all possible.

___________

(Apologizes for the delay in reply 😅 )

graebm: there are ways to test builder changes without doing a full release

you can just install/run builder locally, is the fastest way

or you can get your branch of aws-crt-java to run a branch of aws-crt-builder by changing ci.yml like so:

BUILDER_VERSION: ~v0.9.35~ -> `builder-branch-im-testing`
BUILDER_SOURCE: ~releases~ -> `channels`

But yeah rolling out the change across repos will involve doing a release, and updating which version of builder every other repo is using. But there's also a blessing in doing it that way is that you won't blow up every other repo by accident without going through checkpoints

graebm: my real opposition to this is just the "condition logic with data" aspect, I think that kind of thing is better handled with an explicit script

if you wanted to store a python script in S3 and then have builder pull that down and run it, that would be fine by me

TwistedTwigleg: > But yeah rolling out the change across repos will involve doing a release, and updating which version of builder every other repo is using. But there's also a blessing in doing it that way is that you won't blow up every other repo by accident without going through checkpoints

Good to know! I didn't realize you can just change the builder to point to a branch. That is useful. Maybe then we could just point to a IoT specific branch even and handle it that way, and just keep the branch in sync with main...

And yeah, there is an advantage of having built-in releases, it just seems like it could be slow. But if you can just point to a branch, then maybe it would be fine...

> if you wanted to store a python script in S3 and then have builder pull that down and run it, that would be fine by me

That's a thought I had not considered... Maybe that would be the easiest way to handle something like this...

_________

I can see pros and cons for each. I do see what you are saying about the "condition logic with data" aspect, that makes sense and I can see why we might want to reduce/remove that part.

I will gather up all the possible solutions and put them together, and then look to get everyone's opinion so we can find a solution that everyone agrees on moving forward 👍

TwistedTwigleg:
Closed in favor of #231

PR #227, opened 2023-04-13
Cross build fix

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder/core/data.py
+++ b/builder/core/data.py
@@ -324,7 +324,7 @@ TARGETS = {
             },
         },
 
-        '+cmake_args': [
+        'cmake_args': [
             "-DENABLE_SANITIZERS=ON",
         ],
 
@@ -345,7 +345,7 @@ TARGETS = {
                 ],
             },
         },
-        '!cmake_args': ['-DENABLE_SANITIZERS=ON'],
+        'cmake_args': ['-DENABLE_SANITIZERS=ON'],
         'variables': {
             'exe': '',
         },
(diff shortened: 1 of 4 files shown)


builder/core/project.py
```diff
@@ -26,7 +26,7 @@ def looks_like_code(path):
     return False
 
 
-def _apply_value(obj, key, new_value, apply_before=True):
+def _apply_value(obj, key, new_value, apply_before=False):
```
DmitriyMusatkin, line 29: Note: applying before by default is arguably wrong. we eval config in arch -> host -> target -> compiler -> version order, and we need to apply them in the same order, ex. host sets cmake flag to on, but we want to disable that cmake flag for a certain compiler, then we need to make sure compiler config value comes after host config value (before it was always inserted at the start).

This is just a spot fix and we do need to address how cmake flags are handled in a more generic way

builder/core/data.py
```diff
@@ -550,7 +550,7 @@ class PKG_TOOLS(Enum):
     },
     'gcc': {
         'hosts': ['linux', 'manylinux', 'al2012', 'al2', 'freebsd'],
-        'targets': ['linux', 'freebsd'],
+        'targets': ['linux', 'freebsd', 'android'],
```
DmitriyMusatkin, line 553: note: there was a bug in old versions where it would always use default compiler targets to determine whether cross-compile is supported (and default specified all possible targets as supported). that bug was fixed, but data was not updated to indicate that gcc can cross compile to andorid

JonathanHenson (approved):

waahm7 (approved):

PR #228, opened 2023-04-14
add CMake Flag for gcc 4.8 to disable AVX512-specific optimizations

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder/core/data.py
+++ b/builder/core/data.py
@@ -555,7 +555,10 @@ COMPILERS = {
         'yum_compiler_packages': ['gcc', 'gcc-c++'],
 
         'versions': {
-            '4.8': {},
+            '4.8': {
+                'cmake_args': [
+                    "-DMY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX",
+                ], },
             '5': {},
             '6': {},
             '7': {},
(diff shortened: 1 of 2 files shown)


builder/core/data.py
```diff
@@ -555,7 +555,10 @@ class PKG_TOOLS(Enum):
         'yum_compiler_packages': ['gcc', 'gcc-c++'],
 
         'versions': {
-            '4.8': {},
+            '4.8': {
+                'cmake_args': [
```
DmitriyMusatkin: dont think, this is the right place. we have separate flags for aws-lc https://github.com/awslabs/aws-crt-builder/blob/a31bfacf48d2d10497397674c9928238531b7a50/builder/imports/awslc.py

TingDaoK: but that should be a gcc4.8 only flag, right?

DmitriyMusatkin: it is. this is aws-lc only flag so arguably we should not be passing it to everything that builds with 4.8. 
ideally we would specify specifically for aws-lc that it should build with that flag on 4.8, but lc does not support builder. file i linked above is where aws-lc specific config gets added, we should be able to add compiler specific cmake there

TingDaoK: Yep, that's better, updated

DmitriyMusatkin (approved):

PR #248, opened 2023-08-31
Use CTEST_PARALLEL_LEVEL

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -246,7 +246,7 @@ class CTestRun(Action):
             return
 
         ctest = toolchain.ctest_binary()
-        sh.exec(*toolchain.shell_env, ctest, "--parallel", os.cpu_count(),
+        sh.exec(*toolchain.shell_env, ctest, "--parallel", str(os.cpu_count()),
                 "--output-on-failure", working_dir=project_build_dir, check=True)
         # Try to generate the coverage report. Will be ignored by ctest if no coverage data available.
         sh.exec(*toolchain.shell_env, ctest,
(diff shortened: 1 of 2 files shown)


builder/actions/cmake.py
```diff
@@ -246,7 +246,7 @@ def run(self, env):
             return
 
         ctest = toolchain.ctest_binary()
-        sh.exec(*toolchain.shell_env, ctest,
+        sh.exec(*toolchain.shell_env, ctest, "--parallel", str(os.cpu_count()),
```
bretambrose, line 249: Why?  If this is affecting all repos then that's very concerning.

TingDaoK: the tests are slow. Hmmm, do we have tests that cannot run in parallel? If so, I guess I'll just change s3 to run in parallel instead

bretambrose: I worry about crt-cpp where we have mqtt tests that used unique client ids but share topics and cross-talk between them

bretambrose: Ideally all tests would be truly independent and mutually exclusive, but I don't think we're there yet.

TingDaoK: Sure, I'll just enable it only for aws-c-s3 then.

TingDaoK:
move to just s3

builder/actions/cmake.py
TingDaoK: Still needs to update builder for the script to import environment variables

graebm (approved):

PR #267, opened 2024-01-11
FIPS

TingDaoK:
- Add a helper to import golang 
- install golang>1.18 for ubuntu x64 images, I had a super weird tar extraction failure for those. So, just install them from docker file
- Cmake magic to cross build with FIPS.


builder/imports/golang.py
```diff
@@ -0,0 +1,83 @@
...
+
+import os
+from pathlib import Path
+
+from builder.core.fetch import fetch_and_extract, mirror_package
+from builder.core.project import Import
+import builder.core.util as util
+
+URLs = {
+    'linux-armv6': 'https://go.dev/dl/go1.21.5.linux-armv6l.tar.gz',
```
DmitriyMusatkin, line 12: any specific reason for 1.25.5? 

if we just want latest versions, looks like we can them up here https://go.dev/dl/?mode=json

DmitriyMusatkin (approved):

waahm7 (approved):

.github/docker-images/ubuntu-18-x64/Dockerfile
```diff
@@ -24,6 +24,12 @@ RUN apt-get update -qq \
     ca-certificates \
     && apt-get clean
 
+# Add the longsleep/golang-backports PPA
+RUN apt-get update && apt-get install -y software-properties-common && add-apt-repository ppa:longsleep/golang-backports && apt-get update
```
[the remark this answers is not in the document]
TingDaoK: I did spend half a day to investigate it, and I don't know.

The issue is when we extract the go.tar.gz something failed. I tried different packages, and different ways to extract, no luck.

It only happens for ubuntu, and I gave up

.github/docker-images/alpine-3.16-arm64/Dockerfile
```diff
@@ -18,7 +18,8 @@ RUN apk add \
     ca-certificates \
     cmake \
     bash \
-    aws-cli
+    aws-cli \
+    perl-strictures
```
graebm, line 22: Why do we need this weird specific package?

https://metacpan.org/pod/strictures
> strictures - Turn on strict and make most warnings fatal

Is there something more basic, named like `perl`?

TingDaoK: yeah, but aws-lc needs strict module for perl.

PR #256, opened 2023-10-17
Skip Raspberry Image

xiazhvera:
*Issue #, if available:*

*Description of changes:*
As the raspberry image is not available right now. Always pull the latest image and push to ECR. 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #286, opened 2024-06-18
Docker images: fix raspbian-bullseye, install python3-venv

graebm:
**Issue:**
https://github.com/awslabs/aws-crt-python/pull/573 needs python3's `venv` installed on all docker images, due to [PEP 668](https://peps.python.org/pep-0668/).

While trying to update our `raspbian-bullseye` image, I discovered it's been broken for months. We'd been [hacking around this](https://github.com/awslabs/aws-crt-builder/pull/256) by not even trying to build it anymore, copy/pasting an old working version instead.

**Description of changes:**

- Install `python3-venv` in Docker images that didn't already have it
- Fix `raspbian-bullseye` image.
    - The old image was getting 404 from some repository while running `apt-get update`
    - Use different base image: https://github.com/dtcooper/raspberrypi-os-docker. This was the only easily available `bullseye` image I could find. The implementation looks simple, and it's working.
    - Remove extra stuff from this docker image, like building additional versions of python. I don't know what these were used for (@xiazhvera do you know?). None of our other docker images go through this trouble to build extra python versions. This is already the slowest docker image, due to being armv7, so removing extra work and complication seems extra good.

### DIFF
--- a/.github/docker-images/raspbian-bullseye/Dockerfile
+++ b/.github/docker-images/raspbian-bullseye/Dockerfile
@@ -3,8 +3,6 @@ FROM --platform=linux/arm/v7 dtcooper/raspberrypi-os:bullseye
 ###############################################################################
 # Install prereqs
 ###############################################################################
-# Setup nodejs policy for npm
-RUN apt-cache policy nodejs
 RUN apt-get update -qq \
     && apt-get -y install \
     git \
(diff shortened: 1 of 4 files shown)


waahm7 (approved):

xiazhvera (approved):
I don't remember why we added the various versions of Python. Probably it is for running different Python versions in CI, as we do run python3.7-3.12 for linux (and musl). However, the currently CI did not run against those Python version any more. I think we can remove them safely.

PR #295, opened 2024-09-06
More clang configs

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*


builder/imports/llvm.py
```diff
@@ -53,7 +53,12 @@
 LLVM_VERSION_PATTERNS[10]="-10"
 LLVM_VERSION_PATTERNS[11]="-11"
 LLVM_VERSION_PATTERNS[12]="-12"
-LLVM_VERSION_PATTERNS[13]=""
+LLVM_VERSION_PATTERNS[13]="-13"
+LLVM_VERSION_PATTERNS[14]="-14"
+LLVM_VERSION_PATTERNS[15]="-15"
+LLVM_VERSION_PATTERNS[16]="-16"
+LLVM_VERSION_PATTERNS[17]="-17"
+LLVM_VERSION_PATTERNS[18]="-18"
```
graebm, line 61: In the old code, there's an empty string entry at the end of the array? Do we still need that?
```suggestion
LLVM_VERSION_PATTERNS[13]="-13"
LLVM_VERSION_PATTERNS[14]="-14"
LLVM_VERSION_PATTERNS[15]="-15"
LLVM_VERSION_PATTERNS[16]="-16"
LLVM_VERSION_PATTERNS[17]="-17"
LLVM_VERSION_PATTERNS[18]="-18"
LLVM_VERSION_PATTERNS[19]=""
```

graebm (approved):

builder/imports/llvm.py
DmitriyMusatkin: the way those repos are organized is there a bunch of llvm-toolchain-noble-x folders, with x being the release num, and there is one folder without the number. From what i can tell the numberless folder  points to the latest build of mainline llvm. 

im not a huge fan of the last element of the array pointing to the latest, i.e. before this pr 13th element would have pointed to v19 (which is non stable yet)

PR #312, opened 2024-11-29
lets not pick specific python ver

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder/core/data.py
+++ b/builder/core/data.py
@@ -132,7 +132,7 @@ HOSTS = {
         'pkg_update': 'apt-get -qq update -y',
         'pkg_install': 'apt-get -qq install -y',
         'variables': {
-            'python': "python3.8",
+            'python': "python3",
         },
     },
     'debian': {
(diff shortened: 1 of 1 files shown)


graebm:
hehe, it's PR #312
which is a recent python version

builder/core/data.py
```diff
@@ -132,7 +132,7 @@ class PKG_TOOLS(Enum):
         'pkg_update': 'apt-get -qq update -y',
         'pkg_install': 'apt-get -qq install -y',
         'variables': {
-            'python': "python3.8",
+            'python': "python3",
```
graebm, line 135: Just explaining why this explicitly said "python3.8"...
This was set 2 years ago, in this PR: https://github.com/awslabs/aws-crt-builder/pull/198

back then, on some of our docker images, installing "python3" would get you python 3.6, but you could get something newer by specifying "python3.8"

We've probably stopped using those super old images, so this makes sense.
But do a census to be sure...

There are still a few other places we specify python version, if you want to do a search for "python3."

builder/core/data.py
```diff
@@ -178,7 +178,7 @@ class PKG_TOOLS(Enum):
         'pkg_update': 'zypper refresh && zypper --non-interactive patch',
         'pkg_install': 'zypper install -y',
         'variables': {
-            'python': "python3.9",
+            'python': "python3",
```
graebm, line 181: on `opensuse/leap:15.3` that we're currently using, `python3` == python3.6

maybe we should just be updating the docker image

PR #314, opened 2025-01-06
XCodebuild Environment Variable Setup

xiazhvera:
*Issue #, if available:*

*Description of changes:*
XCodebuild will passing env var "TEST_RUNNER_" into tests. Setup the builder to add "TEST_RUNNER_" to environment variables when use xcodebuild. (https://developer.apple.com/documentation/xcode/environment-variable-reference)

### DIFF
--- a/builder/actions/setup_cross_ci_crt_environment.py
+++ b/builder/actions/setup_cross_ci_crt_environment.py
@@ -29,9 +29,11 @@ class SetupCrossCICrtEnvironment(Action):
         # Kinda silly to have a function for this, but makes the API calls consistent and looks better
         # beside the other functions...
 
-        # XCodebuild will passing env var with "TEST_RUNNER_" prefix into tests. Setup the builder to add
-        # "TEST_RUNNER_" to environment variables when use xcodebuild.
-        # (https://developer.apple.com/documentation/xcode/environment-variable-reference)
+        # We use xcodebuild, Apple's XCode commandline tool, to do iOS/tvOS simulation and tests. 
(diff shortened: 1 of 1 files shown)


builder/actions/setup_cross_ci_crt_environment.py
```diff
@@ -19,10 +19,21 @@
...
+
+    def __init__(self, use_xcodebuild=False):
+        # is_xcodebuild set to true if using Apple XCodebuild
+        self.is_xcodebuild = use_xcodebuild
 
     def _setenv(self, env, env_name, env_data, is_secret=False):
         # Kinda silly to have a function for this, but makes the API calls consistent and looks better
         # beside the other functions...
+
+        # XCodebuild will passing env var with "TEST_RUNNER_" prefix into tests. Setup the builder to add
```
DmitriyMusatkin, line 32: im having hard time understanding what this sentence is saying. 

does excode expect us to pass that variable when we are using it? i what circumstances do we need xcode?

xiazhvera: We use `xcodebuild`, Apple's XCode commandline tool, to do iOS/tvOS simulation and tests. `xcodebuild` would pass in environment variables with prefix `TEST_RUNNER_` for the tests. For example, if we would like to use environment variable `FOO` in the unit test, we would need set `TEST_RUNNER_FOO=BAR`. Then `xcodebuild` will pick up `TEST_RUNNER_FOO` and stripped `TEST_RUNNER_` part, using FOO in it's final unit tests.

DmitriyMusatkin: can you add that info to the comment?

DmitriyMusatkin (approved):

PR #319, opened 2025-01-30
Support: ubuntu-24.04-arm Github runners, Amazon Linux 2023. Fix: aws-crt-cpp sanity test

graebm:
**Issue:**
Github now offers [Linux arm64 runners](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/)! But when I tried using the `ubuntu-24.04-arm` runner in aws-c-common, the unit tests didn't run.

**Description of changes:**
- Remove code that assumed linux/armv8 would **always** be a cross-compile, and therefore shouldn't run tests.
  - There's code in several other places ([1](https://github.com/awslabs/aws-crt-builder/blob/504fa11d0ef9e7df501dc28199ba976ffb3c631d/builder/core/project.py#L611-L628), [2](https://github.com/awslabs/aws-crt-builder/blob/504fa11d0ef9e7df501dc28199ba976ffb3c631d/builder/actions/cmake.py#L236-L243)) that stops test from running when builder is [actually cross-compiling](https://github.com/awslabs/aws-crt-builder/blob/504fa11d0ef9e7df501dc28199ba976ffb3c631d/builder/core/toolchain.py#L93-L95). I could *probably* remove similar overrides that set `run_tests = false` for armv7, armv6, etc in this data.py file ... but I'll keep those as-is for now 

- Add some code to let builder run on Amazon Linux 2023

- Fix aws-crt-cpp sanity-test by setting C++ std version in a different way
    - This took so long to diagnose. TLDR; our tests use custom ASSERT() macros which rely on using the widely supported `##__VA_ARGS__` compiler extension. By default, CMake enables compiler extensions: when you set `CMAKE_CXX_STANDARD=11`, it results in compiler flag `-std=gnu11` (11 with extensions) instead of `-std=c++11` (strict 11). But if you set `export CXXFLAGS=-std=c++11` before building, **new** versions of CMake (but not old versions which is why it still works in some of our CI runs using CMake 3.16.3) will change the default to NOT enable extensions, which breaks our tests. Workaround is to set C++ version via `CMAKE_CXX_STANDARD=11` instead of va `CXXFLAGS=-std=c++11`, which lets CMake continue to use compiler extensions by default. I'll be fixing up our CMake scripts in another PR to force extensions ON when building tests...

### DIFF
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -204,11 +204,11 @@ jobs:
   # Make sure linux compilers + stdlibs are installing properly
   std-compat:
     needs: [package, sanity_test]
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
-        compiler: [gcc-9, clang-9]
+        compiler: [gcc-8, clang-9]
         cxx-std: ["11", "14", "17", "20"]
     steps:
     - uses: aws-actions/configure-aws-credentials@v4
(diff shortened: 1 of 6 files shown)


TingDaoK (approved):
So, we are not modifying every macro now?

PR #341, opened 2025-12-31
Switch cmake_args to ordinary list

azkrishpy:
*Issue #, if available:*

*Description of changes:*
cmake_args being unique list some times causes incorrect behavior since args appended later might not be respected due it the arg already being in the 'unique list'. 
CMake itself processes args from left to right takes the latest value appended, this way args with higher priority can be appended to the end allowing CMake to resolve multiple instances of the same args.

### DIFF
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        host: [ubuntu-22.04, macos-13, macos-14, windows-2025]
+        host: [ubuntu-22.04, macos-latest, windows-2025]
     needs: package
     runs-on: ${{ matrix.host }}
     steps:
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        host: [ubuntu-22.04, macos-latest, windows-2025]
+        host: [ubuntu-22.04, macos-14, macos-latest, windows-2025]
     needs: package
     runs-on: ${{ matrix.host }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/sanity-test.yml
```diff
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        host: [ubuntu-22.04, macos-13, macos-14, windows-2025]
+        host: [ubuntu-22.04, macos-latest, windows-2025]
```
TingDaoK: oh, if it's a sanity-test, maybe try to cover all those are not deprecated?

builder/actions/cmake.py
```diff
@@ -142,7 +142,15 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
             if value:
                 compiler_flags.append(
                     '-DCMAKE_{}_COMPILER={}'.format(opt.upper(), value))
-    cmake_args = UniqueList([
+
+    # Removed UniqueList to make an ordinary list instead. Having multiple arguments in different parts of the command
```
TingDaoK, line 146: i guess the down side is that the order added to the list really matters now (Probably also matter before). But, we don't have a clear doc about when the args will be added to the list here, which will make this hard to use.

.github/workflows/sanity-test.yml
azkrishpy: I think 14 is the earliest version supported now, will have macos-14 and macos-latest maybe.

builder/actions/cmake.py
azkrishpy: yes, order mattered previously as well. But now debugging will be easier since all the flags will be appended irrespective of whether it was added previously.

.github/workflows/sanity-test.yml
TingDaoK: let's use macos-14 and macos-15, latest will keep changing

TingDaoK (approved):

.github/workflows/sanity-test.yml
azkrishpy: Fixed, thanks

PR #342, opened 2026-01-04
Use new method for unique flags

azkrishpy:
*Issue #, if available:*

*Description of changes:*
- Previous use of UniqueList did not allow multiple appends to the list of the same key value pair.
- Adding a new method 'unique_flags' which allows all args to be appended, but in the final set, we remove multiple declarations of certain flags keeping the last occurrence.
- CMake inherently supports multiple arguments and traverses from left to right and constantly updates the flag value based on latest read. However, windows does not support some flags being duplicated ('-A' or '-T') which would cause windows CI to fail.

### DIFF
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -198,7 +198,7 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
     working_dir = env.root_dir if toolchain.cross_compile else os.getcwd()
 
     # Previous use of UniqueList did not allow multiple appends to the list of the same key value pair.
-    # CMake inherently supports multiple arguments and traverses from left to right and constantly updates the flag value based on latest read. 
+    # CMake inherently supports multiple arguments and traverses from left to right and constantly updates the flag value based on latest read.
     # However, windows does not support some flags being duplicated ('-A' or '-T') which would cause windows failed to build.
     # Thus, adding a new method 'unique_flags' to remove multiple declarations of those flags keeping the last occurrence.
     cmake_args = unique_flags(cmake_args, '-A', '-T')
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -197,6 +197,12 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
 
     working_dir = env.root_dir if toolchain.cross_compile else os.getcwd()
 
+    # Previous use of UniqueList did not allow multiple appends to the list of the same key value pair.
+    # Adding a new method 'unique_flags' which allows all args to be appended, but in the final set, we remove
+    # multiple declarations of certain flags keeping the last occurrence.
+    # CMake inherently supports multiple arguments and traverses from left to right and constantly updates the
+    # flag value based on latest read. However, windows does not support some flags being duplicated ('-A'
+    # or '-T') which would cause windows CI to fail.
     cmake_args = unique_flags(cmake_args, '-A', '-T')
 
     # configure
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -148,7 +148,7 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
     # For examp[le, ENABLE_SANITIZERS is set to OFF by default but ON by host config, if we want to turn it off again for
     # specific jobs, having a UniqueList would not allow that.
     # TODO: We need to take into account flags which can have key value pairs while trying to make a unique list and update
-    # keys with the latest 'value' based on priority instead of treating it as an element in the list. This would allow not 
+    # keys with the latest 'value' based on priority instead of treating it as an element in the list. This would allow not
     # duplicating flags.
     cmake_args = [
         "-B{}".format(project_build_dir),
--- a/builder/actions/cmake.py
+++ b/builder/actions/cmake.py
@@ -142,7 +142,15 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
             if value:
                 compiler_flags.append(
                     '-DCMAKE_{}_COMPILER={}'.format(opt.upper(), value))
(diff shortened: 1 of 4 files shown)


builder/actions/cmake.py
```diff
@@ -197,6 +197,8 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
 
     working_dir = env.root_dir if toolchain.cross_compile else os.getcwd()
 
+    cmake_args = unique_flags(cmake_args, '-D', '-A', '-T')
```
TingDaoK: well, we just adding more magic to against the wodoo.🪄
1. where do we get the list? `-d, -a, -t`?
2. Is there anything else is possible to be added?
3. https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-D this `-D` seems to be very complicated to update the cmake cache. Is it right to only keep the last only??

azkrishpy: 1. That is actually magic. I saw only these flags being used in our build commands. Infact, I think windows still allows multiple definitions of -D (and possibly -T) but not -A on windows. And since cmake does not support multiple values for the same flag and it respecting the last found value, I just included -D and -T.
2. possibly.. but I have left it this way for now. It is a dynamic list of args, so any new additions will be a one line change.
3. I did not quite understand the question. But yes, all the -D options that users provide take precedence over the CMakeCache.txt. However the commands we execute are in an isolated environment and while executing, the command cannot respect multiple values for a particular option. That is the only reason I added '-D'. That being said, I am negotiable on what flags need to be added other than -A since -A causes all of windows CIs to fail due to multiple declarations.

TingDaoK: >However the commands we execute are in an isolated environment and while executing, the command cannot respect multiple values for a particular option.

?

TingDaoK: > The option may be repeated for as many CACHE entries as desired.

The documentation says this `-D` can be repeated multiple times??

azkrishpy: Maybe I am misunderstanding something. -D can be repeated multiple times, and we basically write a cache entry for that too. But,
```It's important to note that the order of -C and -D arguments is significant. They will be carried out in the order they are listed, with the last argument taking precedence over the previous ones.```
So only one of the -D values will be respected and that will be the last one right? Like I said, I am open to having multiple entries of -D options since it would help with debugging, I had removed it only because it will get respected only once.

Comment (TingDaoK) [reply builder/actions/cmake.py]: i see, i misunderstood your helper. 
I thought you used -D itself as the prefix to remove the duplicate. 
Until I saw there is a split with `=` to get different keys that starts with the prefix. :)

1. why add this prefix then? Is there anything else doesn't starts with those prefix? Why we don't remove the duplicate for those? It doesn't make sense to me
2. split by the `=` seems to be very fragile. The docs say `-D` can be set from either `-D KEY=VALUE` or `-DKEY=VALUE`. So your `-D KEY` will be different than `-DKEY` even if the key is the same.
3. `-A <platform-name>` that doesn't have a `=` in the whole arg, So, the helper will not remove duplicate of `-A` if the `<platform-name>` is different. Your helper only removes the same `<platform-name>`. I assume windows just doesn't allow multiple `-A` not matter the `<platform-name>` is the same or not

My suggestion is:
1. Either you just try to solve the exact issue that blocks your windows build, with details about why it fails
2. Clearly document what your helper is doing, add some logs to your helper, eg: "`-DAWS_SANITISOR` was set to `True` before, ignore this set to `True`"

File: builder/core/util.py [hunk 2]
```diff
@@ -328,3 +328,23 @@ def append(self, value):
...
+    for arg in reversed(args_list):
+        key = None
+        for prefix in prefixes:
+            if arg.startswith(prefix):
+                key = prefix
+                break
+
+        if key is None or key not in seen:
+            seen.add(key)
+            result.append(arg)
```
TingDaoK, line 348: Add some debug prints if the arg is being ignored.

builder/actions/cmake.py
```diff
@@ -197,6 +197,8 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
 
     working_dir = env.root_dir if toolchain.cross_compile else os.getcwd()
 
+    cmake_args = unique_flags(cmake_args, '-A', '-T')
```
TingDaoK, line 204: add the doc in the code as well to save the step for future to have to dig into the PR description about why this is added.

azkrishpy: Fixed, thanks!

builder/core/util.py
azkrishpy: Fixed, Thanks!

builder/actions/cmake.py
```diff
@@ -197,6 +197,14 @@ def _build_project(env, project, cmake_extra, build_tests=False, args_transforme
 
     working_dir = env.root_dir if toolchain.cross_compile else os.getcwd()
 
+    # Previous use of UniqueList did not allow multiple appends to the list of the same key value pair.
+    # Adding a new method 'unique_flags' which allows all args to be appended, but in the final set, we remove
+    # multiple declarations of certain flags keeping the last occurrence.
+    # CMake inherently supports multiple arguments and traverses from left to right and constantly updates the
+    # flag value based on latest read. However, windows does not support some flags being duplicated ('-A'
+    # or '-T') which would cause windows CI to fail.
```
TingDaoK: Trivial: reword a bit

```suggestion
    # Previous use of UniqueList did not allow multiple appends to the list of the same key value pair.
    # CMake inherently supports multiple arguments and traverses from left to right and constantly updates the flag value based on latest read. 
    # However, windows does not support some flags being duplicated ('-A' or '-T') which would cause windows failed to build.
    # Thus, adding a new method 'unique_flags' to remove multiple declarations of those flags keeping the last occurrence.
```

TingDaoK (approved):
fix and ship

PR #343, opened 2026-02-27
latest-clang

sbSteveK:
Allow CI to specify `latest-clang` and dynamically check llvm for its latest qualification version of clang to build against.

When `latest` is used, we will parse for the latest qualification branch from llvm and use that. Sometimes the latest version that is listed (whether qualification or stable) isn't actually available for the specific codename we are using so we do a check in order of versions we want before selecting and using it.

Also removed hardcoding of versions past 8 to be dynamically set so that we don't need to continually update builder for later version support. Left comments in case it breaks and we need to return to hardcoding but that shouldn't be necessary.

### DIFF
--- a/builder/core/toolchain.py
+++ b/builder/core/toolchain.py
@@ -344,14 +344,7 @@ class Toolchain(object):
 
     @staticmethod
     def is_compiler_installed(compiler, version):
-        """ Returns True if the specified compiler is already installed, False otherwise.
-        For 'latest' version of 'clang' compiler, this will resolve the actual version first. """
-        if version == 'latest' and compiler == 'clang':
-            # Import here to avoid circular dependency
-            from builder.imports.llvm import LLVM
-            resolved_version = LLVM.resolve_latest_version()
-            if resolved_version:
-                version = resolved_version
+        """ Returns True if the specified compiler is already installed, False otherwise """
         compiler_path, found_version = Toolchain.find_compiler(
             compiler, version)
         return compiler_path != None
--- a/builder/core/toolchain.py
+++ b/builder/core/toolchain.py
@@ -182,8 +182,16 @@ class Toolchain(object):
 
     @staticmethod
     def find_llvm_tool(name, version=None):
-        """ Finds clang, clang-tidy, lld, etc at a specific version, or the
-        latest one available. If version is 'latest', resolves it dynamically. """
+        """ Finds clang, clang-tidy, lld, etc at a specific version.
+
+        - If version is None (default): searches all known/hardcoded LLVM versions
+          (from COMPILERS data), highest first, and returns the first one installed
+          on the system. No network calls are made.
+        - If version is 'latest': dynamically resolves the newest available LLVM
+          version from apt.llvm.org (requires a network call), then searches for
+          that version. Falls back to the default (None) behavior if resolution fails.
+        - If version is a specific string (e.g. '18'): searches only for that version.
+        """
         if version == 'latest':
             # Import here to avoid circular dependency
             from builder.imports.llvm import LLVM
--- a/builder/core/toolchain.py
+++ b/builder/core/toolchain.py
@@ -36,6 +36,9 @@ def _compiler_version(cc):
 
 
 def _find_compiler_tool(name, versions):
+    # Filter out 'latest' from versions list. It will be handled separately.
+    versions = [v for v in versions if v != 'latest']
+
     # look for the default tool, and see if the version is in the search set
     path = util.where(name, resolve_symlinks=False)
     if path:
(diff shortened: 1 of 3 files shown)


builder/core/toolchain.py
```diff
@@ -329,7 +345,14 @@ def _find_compiler():
...
     def is_compiler_installed(compiler, version):
-        """ Returns True if the specified compiler is already installed, False otherwise """
+        """ Returns True if the specified compiler is already installed, False otherwise.
+        For 'latest' version of 'clang' compiler, this will resolve the actual version first. """
+        if version == 'latest' and compiler == 'clang':
+            # Import here to avoid circular dependency
+            from builder.imports.llvm import LLVM
+            resolved_version = LLVM.resolve_latest_version()
+            if resolved_version:
+                version = resolved_version
```
TingDaoK: this is redundant, `find_compiler` will call `find_llvm_tool` and you handle the latest verison tag again over there. It will be easier to follow if we can keep the handling of the special version in one place.

builder/core/toolchain.py
```diff
@@ -228,7 +243,8 @@ def _find_msvc(version, install_vswhere=True):
 
     @staticmethod
     def find_compiler(compiler, version=None):
-        """ Returns path, found_version for the requested compiler if it is installed """
+        """ Returns path, found_version for the requested compiler if it is installed.
+        If version is 'latest' for clang, it will be resolved dynamically. """
```
TingDaoK: It seems to be confusing to have this comment but no special handling of the `latest` tag happening in here.

builder/core/toolchain.py
```diff
@@ -180,8 +183,20 @@ def find_gcc_tool(name, version=None):
     @staticmethod
     def find_llvm_tool(name, version=None):
         """ Finds clang, clang-tidy, lld, etc at a specific version, or the
-        latest one available """
-        versions = [version] if version else _clang_versions()
+        latest one available. If version is 'latest', resolves it dynamically. """
```
TingDaoK: So, we are trying to resolve the latest one available when the version is not set, but we are also introducing a new `latest` version? 
Sounds like we should just resolve the latest version dynamically when the version is `None` instead of the latest one in the hard-coded list?

builder/imports/llvm.py
```diff
@@ -44,50 +48,52 @@
...
-    exit 3
-fi
+# Determine the version string suffix for package names
+# Old versions (3, 6, 7) have special suffixes and are explicitly set to hardcoded values.
+# 8+ use -VERSION format. This should be the case moving forward but may need to be changed
+# if llvm changes it in the future and `latest` fails.
+case "$LLVM_VERSION" in
+    3)  LLVM_VERSION_STRING="-3.9" ;;
+    6)  LLVM_VERSION_STRING="-6.0" ;;
+    7)  LLVM_VERSION_STRING="-7" ;;
```
bretambrose: Why is this line needed?

builder/imports/llvm.py
```diff
@@ -10,11 +10,15 @@
...
+import urllib.request
 
-# this is a copy of https://apt.llvm.org/llvm.sh modified to add support back in
-# for older versions of clang < 8, and removed the need for clangd, lldb
+# This is a modified version of https://apt.llvm.org/llvm.sh that:
+# - Supports older versions of clang (3.9, 6.0, 7)
+# - Supports any future version (8+) dynamically
+# - Removes the need for clangd, lldb
 
 LLVM_SH = """\
```
bretambrose, line 23: Why do we do this rather than a separate file?

sbSteveK: Unsure, I touched as little of the existing framework as I could because I didn't want to break anything else. A lot of this felt pretty hacky. If you mean the entire implementation as a whole, it does still feel like the changes should be in this file since we're parsing the arguments for clang here.

builder/imports/llvm.py
sbSteveK: missed it. Removed.

builder/imports/llvm.py
TingDaoK: have you tried the latest https://apt.llvm.org/llvm.sh?
Look into this, it seems to be outdated comparing to what I just downloaded. 

Does the latest script supports future version (8+)?

builder/core/toolchain.py
sbSteveK: Removed comment about `latest`

builder/imports/llvm.py
sbSteveK: Two immediate problems are that the latest doesn't have immediate support for anything lower than clang-9 (so we'd need to hardcode earlier versions) and it has its version hardcoded up to 21 so later versions as things change will need to be manually addressed which is what I'm trying avoid doing here. We might be able to use this as a NEW base to work off but I don't see a strong reason why we should.

builder/core/toolchain.py
sbSteveK: I could have sworn I needed to do this for something but looking it through I can't figure out why I did it anymore.... Curse of the weekend.

builder/core/toolchain.py
sbSteveK: We introduce `latest` to be explicit about what we want to do when calling from CI; find the latest version of clang that can be used from llvm. `None` simply asks to use any available version and doesn't go through the process of looking for latest from llvm. I think we should keep this separation in behavior.

TingDaoK: at least update the comments, this is confusing to say find the latest one available, then if the version is latest, get another latest one dynamically.

TingDaoK (approved):

PR #347, opened 2026-04-27
ubuntu-18-x64 python3 symlink

sbSteveK:
In ubuntu-18-x64 we build Python 3.10 from source and symlink it to python but not python3. This change also symlinks to python3 so that downstream CI tests where scripts try to use python3 don't fail to find pip and/or run into other python3 command related issues.


sfod (approved):

DmitriyMusatkin:
can we kill it at some point? have ubuntu min that all repos use and have it point to 20 or something?

PR #353, opened 2026-06-22
Add CI to check ABI compliance

azkrishpy:
## Add ABI compliance check GitHub Action for CRT C libraries

*Description of changes:*

Adds a reusable composite action (`.github/actions/check-abi`) that dumps and compares the ABI/API of a CRT C shared library between a PR's head and its base ref, and labels the PR with the implied semver bump -- `patch`, `minor`, or `needs-review` when a real break slips in unlabeled as such. Details of working added to `README.md` for the action.

### Why

CRT C libraries (`aws-c-common`, `aws-c-io`, `aws-c-s3`, …) ship as shared libraries that downstream consumers link against dynamically. A backward-incompatible change (a removed/renamed symbol, an altered struct layout, a changed signature) silently breaks those consumers at runtime with no compile-time or load-time warning. This action moves detection to PR time, where the author has the context, instead of leaving it to be discovered downstream.

### Structure

- `action.yml` -- pulls the image, runs the in-container orchestrator, labels the PR.
- `scripts/abi_check.sh` -- in-container orchestrator (build -> dump -> compare -> report -> gate).
- `scripts/build.sh` -- resolve base ref, create worktree, build both refs (parallel).
- `scripts/dump.sh` -- locate both `.so`s, dump their ABI (parallel).
- `scripts/compare.sh` -- run the compliance check (binary + source), record rc/pct.
- `scripts/check_constants.sh` -- ctags-based supplement for the removed-constant blind spot.
- `scripts/report.py` -- sanitize and write the verdict + report body to the job summary.
- `scripts/gate.sh` -- parse each report's structured verdict, choose the patch/minor/needs-review label, fail only on tool error.
- `README.md` -- how it works, usage, inputs, scope/limitations, and label semantics.

### Notes / limitations

- The consumer workflow must checkout with `fetch-depth: 0` so the base branch history is available for the worktree and `git merge-base`, and grant `pull-requests: write` to label the PR (skipped automatically, non-fatally, on fork PRs where the token is forced read-only).
- Linux (x86_64) API only -- platform-guarded code (`#if defined(_WIN32)`/Apple-only) is never compiled or checked; a deliberate scope decision, not an oversight.
- No submodule support (both refs build from a `git worktree`); a non-issue for the `aws-c-*` libraries today but not a drop-in fit for a repo like `aws-crt-cpp`.
- Single-library scope: each run diffs one library's own ABI/API against its own previous version. It does not detect a break that only manifests when a *different* library in the dependency graph is upgraded without a matching redeploy -- that must be caught by running this check in the dependency's own repo. This has been researched and is treated as an accepted, expected limitation given each CRT package versions independently, not a tooling gap.
- `libabigail` (`abidw`/`abidiff`) was evaluated as an alternative and moved back to `abi-compliance-checker` + `abi-dumper`. The deciding factor was source (API) compatibility, not just binary (ABI) compatibility: `abi-compliance-checker` can run in `-binary -source` mode and separately reports breaks that are 100% binary-compatible but fail to *recompile* (e.g. a renamed public enum constant) -- a class of break libabigail's tooling does not surface with the same fidelity. Since a source break is a strictly worse failure mode for a caller than a binary-only one, and this action needs to flag both distinctly (see the three-way label above), abicc's dual-mode reporting made it the better fit despite libabigail being the more actively maintained project overall. This might change and we might have to move the tool we use in which case, this might be a good starting point: https://github.com/awslabs/aws-crt-builder/pull/359.


.github/actions/check-abi/action.yml
```diff
@@ -0,0 +1,82 @@
...
+    required: false
+    default: 'https://d19elf31gohf1l.cloudfront.net'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Install ABI tools
+      shell: bash
+      run: |
+        sudo apt-get update -qq
```
sfod: `apt-get` assumes we verify ABI compatibility on Linux only. Probably it's fine for all repos except aws-c-io, which has [windows-specific](https://github.com/awslabs/aws-c-io/blob/v0.27.2/include/aws/io/tls_channel_handler.h#L224-L230) and [apple-specific](https://github.com/awslabs/aws-c-io/blob/v0.27.2/include/aws/io/tls_channel_handler.h#L240) public API.

sfod: I'd also document the requirements for the CI job in a consumer - currently it's Ubuntu, but handling aws-c-io might add another stuff.

.github/actions/check-abi/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Usage
+
+Add a workflow to the library repo (see `aws-c-s3/.github/workflows/check-abi.yml`):
+
+```yaml
+- name: Check ABI
+  # note: using "@main" because "@${{env.BUILDER_VERSION}}" doesn't work
+  # https://github.com/actions/runner/issues/480
+  uses: awslabs/aws-crt-builder/.github/actions/check-abi@main
```
sfod: Using `@main` is a risky thing. Can we introduce a `latest` tag pointing to the latest release?

.github/actions/check-abi/README.md
```diff
@@ -0,0 +1,66 @@
+# check-abi
+
+Dump and compare the ABI of a CRT C shared library between a pull request's head
+and its base ref, and fail the build on an unauthorized binary-incompatible
+change. This is intended for use with the CRT C libraries (ex: aws-c-s3, aws-c-io).
+
+## What it does
+
+1. **Install ABI tools** — `abi-dumper`, `abi-compliance-checker`,
+   `universal-ctags`, plus build dependencies (`cmake`, `g++`, `libssl-dev`).
```
sfod: Are there any alternatives to these tools? e.g. https://github.com/lvc/abi-compliance-checker wasn't updated for 5 years (release is 8 yo). `homebrew-core` [raised the same concerns](https://github.com/Homebrew/homebrew-core/pull/173741) and decided to deprecate these tools.

.github/actions/check-abi/scripts/build.sh
```diff
@@ -0,0 +1,102 @@
...
+
+# --- Build both refs in parallel ---------------------------------------------
+# builder installs to <source_dir>/build/install. The same builder.pyz builds
+# both refs; each invocation is an independent process with its own source dir.
+build_ref() {
+  local src_dir="$1"
+  ( cd "$src_dir" && python3 "$BUILDER_PYZ" build -p "$LIB_NAME" \
+      --cmake-extra=-DBUILD_SHARED_LIBS=ON \
+      --cmake-extra=-DBUILD_TESTING=OFF \
+      --cmake-extra="-DCMAKE_C_FLAGS=-g -Og" \
```
sfod: There is a good chance these options will be overwritten by [`--config RelWithDebInfo` passed by default](https://github.com/awslabs/aws-crt-builder/blob/v0.9.96/builder/actions/cmake.py#L210-L211). I'd verify this.

.github/actions/check-abi/scripts/report.py
```diff
@@ -0,0 +1,104 @@
...
+        return ''
+
+    content_start = search_from + m.end()
+    if end < content_start:
+        return ''
+    return html[content_start:end]
+
+
+def main():
+    summary_path = _env('GITHUB_STEP_SUMMARY')
```
sfod, line 164: This was flagged as a low-security issue by my agent: we write the raw output of a 3p tool to a file that github then renders.

.github/actions/check-abi/scripts/build.sh
```diff
@@ -0,0 +1,102 @@
...
+
+# --- Create an isolated worktree for the base ref ----------------------------
+BASE_TMPDIR="$(mktemp -d)" || { echo "ERROR: mktemp -d failed" >&2; exit 1; }
+BASE_WORKTREE="${BASE_TMPDIR}/worktree"
+
+# Record the tmp dir immediately so the cleanup step can remove it even if the
+# worktree add or a later build fails.
+echo "ABI_BASE_TMPDIR=${BASE_TMPDIR}" >> "$GITHUB_ENV"
+
+if ! git -C "$HEAD_DIR" worktree add "$BASE_WORKTREE" "$BASE_REF"; then
```
sfod, line 66: `git worktree` doesn't populate submodules. This CI jobs is supposed to work with aws-c-* libs, so maybe it's not an issue, just wanted to make sure we're not intended to use it with, say, aws-crt-cpp.

azkrishpy: yes compliance check will be done only at the native level.

.github/actions/check-abi/scripts/report.py
azkrishpy: ah! fair enough.. is there an alternative? I thought the summary being visible on the action itself would be the simplest way to view what changes are affecting ABI.

.github/actions/check-abi/scripts/build.sh
azkrishpy: I will take a closer look into this but from what I remember in a previous investigation, cmake takes the same flag multiple times and is happy to ignore every set value up until the last one. and since cmake-extra is [appended to the end](https://github.com/awslabs/aws-crt-builder/blob/e144d9bacd44ecb5698a770fdeec61fb58d193f4/builder/actions/cmake.py#L168) (atleast all of the variables we have added here) it should be fine..?

.github/actions/check-abi/README.md
azkrishpy: I agree with the concern, however the only other alternative I found was [libabigail](https://sourceware.org/libabigail/) which had a lot more dependencies. abicc was also something other sdks used previously for checking ABI breaks. Since it was catching all the ABI breaks we could think off and the workflow itself was to ensure the CI provides the developer an idea on whether the PR breaks ABI we proceeded with abicc.

.github/actions/check-abi/action.yml
azkrishpy: This is actually up for discussion. Since most API is intended to be platform agnostic, the intention itself was to only to ABI checks on linux alone.

.github/actions/check-abi/action.yml
```diff
@@ -0,0 +1,82 @@
...
+        sudo apt-get install -y --no-install-recommends \
+          cmake g++ libssl-dev abi-compliance-checker abi-dumper universal-ctags
```
DmitriyMusatkin: why do we need libssldev fo?

.github/actions/check-abi/action.yml
```diff
@@ -0,0 +1,82 @@
...
+    default: 'releases'
+  builder-host:
+    description: 'Builder artifact host URL.'
+    required: false
+    default: 'https://d19elf31gohf1l.cloudfront.net'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Install ABI tools
```
DmitriyMusatkin: worth considering just having an image with all the tools we need prebuild?

azkrishpy: definitely. will look into how we do this already and incorporate something similar.

.github/actions/check-abi/action.yml
azkrishpy: iirc this was not building against aws-lc and ubuntu needs libssl-dev for openssl? (I built this based on an internet search, will remove it and see if everything still goes fine)

.github/actions/check-abi/README.md
DmitriyMusatkin: what do you mean by libabigail has more deps? seems like available in dnf on latest ubuntu.
i dont mind abi checker but eventually we will have to consider changing if they deprecate it

sfod: Does `libabigail` provide all the required functionality? If it's maintained and can replace abi-* tools, extra dependencies don't seem such a big deal.
I think it'd worth spending some time on additional research for alternatives. If there's nothing then yeah, we can stick with abi-dumper and abi-compliance-checker.

azkrishpy: @DmitriyMusatkin By more deps I meant abigail-tools itself has a bunch of [dependencies](https://sourceware.org/libabigail/#:~:text=autoconf,python%2Dsphinx) that I am not sure they maintain. That said, since this has become a popular demand I am attempting to do the same thing using libabigail to see how well that works. if it is working the same I will make the changes to update the tool we use to libabigail.

.github/actions/check-abi/scripts/report.py
azkrishpy: added some sanitization of inputs from the 3p tool.

.github/actions/check-abi/action.yml
azkrishpy: removed libssl since we don't build against openssl and libssl is provided in the env by default if necessary.

.github/actions/check-abi/README.md
azkrishpy: Per previous discussion an attempt was made to switch the tool used to libabigal in [this pr](https://github.com/awslabs/aws-crt-builder/pull/359). It generated a [similar report](https://github.com/awslabs/aws-c-common/actions/runs/29474425173) but there was no source compatibility check which usually is an added layer of protection against backward incompatible changes. abicc does a best effort check for source compat and an exception that was noticed was that it does not check public constants so added a script that helps with checking for missing public enum constants or macros. Comments in the script explains in more detail. I prefer using the current tool since it is stable and more widely used.. as far as support is concerned libabigail has also not received any contributions in the past 4 years so I believe they are equally on track to be fully deprecated.

.github/actions/check-abi/action.yml
azkrishpy: added a new docker image with all the tools required and running tests from that container.

.github/actions/check-abi/scripts/build.sh
azkrishpy: Yeah CMAKE_C_FLAGS was getting overridden. update to `CMAKE_C_FLAGS_RELWITHDEBINFO`

.github/actions/check-abi/action.yml
```diff
@@ -0,0 +1,200 @@
...
+      # process (they chain state through an in-container GITHUB_ENV). The gate
+      # exits non-zero ONLY when abi-compliance-checker could not produce a
+      # verdict (rc >= 2); a compatible (rc 0) or incompatible (rc 1) ABI both
+      # succeed here and are surfaced as a PR label by the next step.
+      #
+      # The verdict escapes the container via stdout, not a bind-mounted file:
+      # `docker run` already streams the container's stdout back to this host
+      # process, so gate.sh just prints a marker line ("ABI_LABEL_RESULT::<label>")
+      # and we grep it out of the captured output below. A host-mounted file
+      # here was solving a problem that doesn't exist -- and it introduced a
```
sfod: Looks like an AI summary of an intermediate problem you encountered. I don't see any value in such comments. This is not the only comment, so the whole PR needs additional cleaning up.

.github/actions/check-abi/scripts/gate.sh
```diff
@@ -0,0 +1,166 @@
...
+# This check is informational, not pass/fail: every real verdict (patch,
+# minor, or needs-review) succeeds -- the job fails only when the check itself
+# could not run (so there is no trustworthy verdict to label with).
+#
+# LABEL DECISION -- three-way, not two-way. abicc's own exit code / overall
+# "verdict:compatible|incompatible" field is NOT used to choose the label,
+# because it conflates two things a reviewer needs to see separately:
+#   - abicc flags plenty of harmless changes at Low severity (a pure
+#     parameter-name rename, an enum-member rename with a value that doesn't
+#     change) as "incompatible" -- confirmed empirically against the real
```
sfod: `compare.sh` contains the following:
>  -strict : treat low-severity issues as problems too.

But here you have a whole section describing how to ignore low-severity issues.
Won't it be easier to remove `-strict` option?

DmitriyMusatkin (approved):

.github/actions/check-abi/scripts/report.py
```diff
@@ -0,0 +1,225 @@
...
+    if not summary_path:
+        print('GITHUB_STEP_SUMMARY not set; skipping summary publish')
+        return 0
+
+    lib_name = _env('ABI_LIB_NAME', '(unknown)')
+    pct = _env('ABI_PCT', '?')
+    src_pct = _env('ABI_SRC_PCT', '?')
+    rc_raw = _env('ABI_RC')
+    report_html = _env('ABI_REPORT_HTML')
+    src_report_html = _env('ABI_SRC_REPORT_HTML')
```
sfod, line 174: When `compare.sh` doesn't run at all (e.g. `build.sh` fails), this leads to `rc` set to -1  and `report` vars being empty, which leads to `_compute_label` returning patch.

.github/actions/check-abi/README.md
```diff
@@ -0,0 +1,172 @@
...
+| `github-token` | no | `github.token` | Token used to label the PR (needs `pull-requests: write`). |
+| `patch-label` | no | `patch` | Label applied when the ABI is backward-compatible. |
+| `minor-label` | no | `minor` | Label applied when the ABI changed. |
+| `base-ref` | no | _(none)_ | Explicit base ref to diff against (e.g. a previous release tag), overriding the PR base ref. Only for non-PR callers (e.g. a release workflow); leave unset in PR workflows so the PR gets labeled. |
+
+## The label
+
+The check maps the abi-compliance-checker result to a semver label:
+
+| abicc exit | Meaning | Result |
```
sfod: `need-review` should be added to the table.

.github/actions/check-abi/action.yml
azkrishpy: left everything as it is mostly for reference.. but yeah I can always look up previous commits in the PR so will clean this up.

sfod (approved):

PR #354, opened 2026-06-25
Loosen compiler version restriction

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Loosens the compiler version validation so that system-detected compiler versions no longer need to exactly match a version listed in `data.py`. Instead, any version at or above the minimum supported version is accepted.

### DIFF
--- a/.github/workflows/sanity-test.yml
+++ b/.github/workflows/sanity-test.yml
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        host: [ubuntu-22.04, macos-14, macos-15, windows-2025]
+        host: [ubuntu-22.04, macos-14, macos-15, macos-26, windows-2025]
     needs: package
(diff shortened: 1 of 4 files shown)


sfod (commented):
Can we add a few CI jobs verifying the changes? For example, `macOS with apple clang 17`, `latest macOS with apple clang default`, maybe something else.

xiazhvera:
Updated to add tests for the default compiler detection: 

If we use default complier, the builder will use the system installed compiler as long as the system compiler is greater than the minimal compiler version: https://github.com/awslabs/aws-crt-builder/actions/runs/28392972696/job/84124499569?pr=354
If we explicitly claim a complier version, and it is not in the list. The builder failed: https://github.com/awslabs/aws-crt-builder/actions/runs/28392972696/job/84124499412?pr=354

sfod (approved):

PR #362, opened 2026-07-27
Updating python version

rakshil14-2:
*Issue #, if available:*

*Description of changes:*
- Updating Python version to python 10 


.github/docker-images/ubuntu-20-x64/Dockerfile
```diff
@@ -26,12 +26,43 @@ RUN apt-get update && apt-get install -y software-properties-common && add-apt-r
 # Install Go from the PPA
 RUN apt-get install -y golang-go
 
+# Install dependencies for building Python from source
+RUN apt-get update \
```
azkrishpy, line 33: are we sure we need every dependency here

azkrishpy: did a quick search seems like software-properties-common, xz-utils (no need of the cmd line tool since we have liblzma-dev). if these are recommended by python or something we might as well link the recommendation here to avoid confusion.

azkrishpy (commented):
Maybe a one-liner explaining why we are moving from default version to 3.10?

.github/docker-images/ubuntu-20-x64/Dockerfile
rakshil14-2: These are the CPython "Suggested build environment" packages for Debian/Ubuntu:
https://github.com/pyenv/pyenv/wiki#suggested-build-environment. 
Dropping : xz-utils and software-properties-common.

sfod (approved):

azkrishpy (approved):

PR #363, opened 2026-07-28
Bake ubuntu-toolchain-r ppa into ubuntu

sfod:
*Issue #, if available:*
Currently, `apt-add-repository -y ppa:ubuntu-toolchain-r/test` runs each time an Ubuntu image is used in CI (it provides newer versions of gcc). It seems we started hitting launchpad API limits because multiple CI jobs are executing it in parallel.

*Description of changes:*
Bake ubuntu-toolchain-r PPA into ubuntu images. The PPA itself (ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu) is served by a CDN and should be reliable, so compilers are not added into the images.

`apt-add-repository -y ppa:ubuntu-toolchain-r/test'` will still be run in runtime if aws-crt-builder is executed on Ubuntu without this PPA, i.e. not on one of the images provided by it.

### DIFF
--- a/.github/docker-images/ubuntu-18-x64/Dockerfile
+++ b/.github/docker-images/ubuntu-18-x64/Dockerfile
@@ -16,6 +16,9 @@ RUN apt-get update -qq \
     ca-certificates \
     && apt-get clean
 
+# ubuntu-toolchain-r PPA provides newer gcc versions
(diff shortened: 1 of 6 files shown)


TingDaoK (approved):
❤️

PR #372, opened 2026-08-10
Dynamically Adding Node version

rakshil14-2:
*Description of changes:*

- Based on the Node version present in builder.json of repo a Node class corresponding to that version will be dynamically created
- At first in `project.py` a node class with a particular version is looked up. If not found the class is created 

### DIFF
--- a/builder/imports/nodejs.py
+++ b/builder/imports/nodejs.py
@@ -156,9 +156,7 @@ _NODE_VERSION_RE = re.compile(r'^node(?:js)?-?(\d+)$')
 
 
 def find_node(name):
-    """Dynamically create a NodeJS import for any requested node version.
-
-    """
+    """Dynamically create a NodeJS import for any requested node version."""
     match = _NODE_VERSION_RE.match(name.lower())
     if not match:
         return None
--- a/builder/imports/nodejs.py
+++ b/builder/imports/nodejs.py
@@ -4,7 +4,7 @@
 
 from builder.core.fetch import fetch_script, fetch_and_extract
 from builder.core.host import current_os, current_arch
-from builder.core.project import Import
+from builder.core.project import Import, Project
 import builder.core.util as util
 from builder.actions.install import InstallPackages
 from builder.actions.script import Script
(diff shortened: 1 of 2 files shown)


azkrishpy:
What is the end goal? We don't seem to be adding new docker images like: https://github.com/awslabs/aws-crt-builder/blob/d1b878d13283d10741accc82201686bd18a37220/.github/workflows/create-release.yml#L89
or changing anything within builder. If there are more changes, you can add it within this PR, if not provide some context in the description as to where you would be using it.

azkrishpy (approved):

xiazhvera (approved):
The change looks straightforward, while I have the same question as Krish did. What is the end goal? 

How did we plan to use the new versions? Is it necessary to create a class for each version? Could we just pass in the versions instead? Should we change the default node version?

builder/imports/nodejs.py
```diff
@@ -150,21 +150,25 @@ def normalize_version(v):
...
+# Matches import names like "node-20", "node20", "nodejs-22", etc.
+# The base "nodejs" import (no digits) is handled by the NodeJS class itself.
+_NODE_VERSION_RE = re.compile(r'^node(?:js)?-?(\d+)$')
 
 
-class Node14(NodeJS):
-    def __init__(self, **kwargs):
-        super().__init__(version='14', **kwargs)
+def find_node(name):
+    """Dynamically create a NodeJS import for any requested node version.
```
azkrishpy: ?

builder/imports/nodejs.py
```diff
@@ -150,21 +150,25 @@ def normalize_version(v):
...
 
+    """
```
azkrishpy:
```suggestion
def find_node(name):
    """Dynamically create a NodeJS import for any requested node version."""
    match = _NODE_VERSION_RE.match(name.lower())
    if not match:
        return None
    version = match.group(1)
```

azkrishpy (commented):
Might be worth updating the PR title and description with what has changed

xiazhvera:
Looks good to me. While it seems the CI was not triggered properly.

PR #380, opened 2026-09-02
Bake llvm gpg key into ubuntu Docker images

sfod:
Fix intermittent CI failures:
```                                                                                   
+ curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key                           
+ apt-key add -                                                                  
curl: (7) Couldn't connect to server                                             
gpg: no valid OpenPGP data found. 
```

by baking apt.llvm.org signing key into the Ubuntu 18/20/22 CI images.

### DIFF
--- a/builder/imports/llvm.py
+++ b/builder/imports/llvm.py
@@ -98,7 +98,10 @@ esac
 
 # install everything
 if [[ $LLVM_VERSION -ne 3 ]]; then
-    curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+    # Only fetch the key here if it isn't already present in the system.
+    if [ ! -f /etc/apt/trusted.gpg.d/apt.llvm.org.asc ]; then
+        curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+    fi
     add-apt-repository "${REPO_NAME}"
     apt-get update
 fi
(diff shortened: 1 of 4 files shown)


builder/imports/llvm.py
```diff
@@ -98,7 +98,10 @@
 
 # install everything
 if [[ $LLVM_VERSION -ne 3 ]]; then
-    curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+    # Only fetch the key here if it isn't already present in the system.
+    if [ ! -f /etc/apt/trusted.gpg.d/apt.llvm.org.asc ]; then
+        curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
```
bretambrose: So all `apt-key add` does is write it to `/etc/apt/trusted.gpg.d/apt.llvm.org.asc` ?

bretambrose (approved):

builder/imports/llvm.py
sfod: `apt-key` is a legacy way to add keys and as far as I know it installs them in a binary format to a different dir (but from functionality perspective, the asc and binary keys act the same).
I kept `apt-key` to reduce changes, but I'm reading more about this and it turns out `apt-key` was removed in Ubuntu 24. So, it's time to get rid of it.

PR #381, opened 2026-09-09
add raspbian-bookworm

sbiscigl:
*Description of changes:*

[raspbian-bullseye builds have started failing](https://github.com/awslabs/aws-crt-cpp/actions/runs/34276369620/job/102265462044?pr=905) because [bullseye has EOL-ed](https://www.debian.org/News/2026/20260831). this creates a bookworm based image, removing the old image because it will no longer work as expected.


.github/docker-images/raspbian-bookworm/Dockerfile
```diff
@@ -26,8 +26,8 @@ RUN apt-get update -qq \
 ###############################################################################
 # Python/AWS CLI
 ###############################################################################
-RUN python3 -m pip install --upgrade setuptools virtualenv \
-    && python3 -m pip install --upgrade awscli \
+RUN python3 -m pip install --break-system-packages --upgrade setuptools virtualenv \
```
DmitriyMusatkin, line 29: why?

.github/docker-images/raspbian-bookworm/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apt-get update -qq \
     clang \
     cmake \
     # Python
-    python3.9 \
```
DmitriyMusatkin, line 15: having random python scares me a bit. we try to do lowest supported

.github/docker-images/raspbian-bookworm/Dockerfile
sbiscigl: this version of debian ships with a python already in the image, that errors if you try to pip install as its managed by apt, and this lets you get past that.

.github/docker-images/raspbian-bookworm/Dockerfile
sbiscigl: debian ships with 3.11, so i was just err-ing on the side of using what came in the image, what is the lowest supported?

sbiscigl: pinned to 3.11

.github/docker-images/raspbian-bookworm/Dockerfile
sfod: I did a quick research: this flag is needed here only because debian (raspberry is debian-based) adopted [PEP-668](https://peps.python.org/pep-0668/), which marks the system’s Python installation as an externally managed environment. We're installing the same python packages via pip in other linux-based images without breaking things. So here it also shouldn't cause any issues.

sfod (approved):

DmitriyMusatkin (approved):

PR #382, opened 2026-09-11
Add latest tag for images

sfod:
Add per-OS-family latest image aliases (e.g. aws-crt-ubuntu-latest-x64) that track the newest member of each OS family, so downstream consumers can reference an OS family without pinning a specific version. Single-member OS families (raspbian, fedora, etc.) get an alias too.
                                                                      
Also deduplicates create-release.yml and create-channel.yml.

## New `latest` aliases
                                                                      
| Alias | Source |
|---|---|
| `ubuntu-latest-x64` | `ubuntu-22-x64` |
| `ubuntu-latest-aarch64` | `ubuntu-20-aarch64` |
| `al-latest-x64` | `al2023-x64` |
| `manylinux-latest-x86` / `-x64` / `-aarch64` | `manylinux_2_28-*` |
| `musllinux-latest-x64` / `-aarch64` | `musllinux-1-1-*` |
| `alpine-latest-x64` / `-x86` / `-arm64` / `-armv7` / `-armv6` | `alpine-3.16-*` |
| `rhel-latest-x64` | `rhel8-x64` |
| `fedora-latest-x64` | `fedora-34-x64` |
| `opensuse-latest` | `opensuse-leap` |
| `openwrt-latest-x64` | `openwrt-x64-openjdk8` |
| `raspbian-latest` | `raspbian-bookworm` |


.github/workflows/build-images.yml
```diff
@@ -0,0 +1,182 @@
...
+        variant:
+        - name: manylinux1-x86
+        - name: manylinux1-x64
+        - name: manylinux2014-x86
+        - name: manylinux2014-x64
+        - name: manylinux2014-aarch64
+        - name: manylinux_2_28-x86
+          latest-alias: manylinux-latest-x86
+        - name: manylinux_2_28-x64
+          latest-alias: manylinux-latest-x64
```
xiazhvera, line 55: Trivial: Just saying it's a bit annoying that we have to manually mark each latest image. I couldn't think of a better solution though.

xiazhvera (approved):

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,182 @@
...
+        # consumers can always reference `<family>-latest-<arch>` without knowing the current version.
+        # The abi/msan variants are intentionally excluded since they are specialized CI images.
+        variant:
+        - name: manylinux1-x86
+        - name: manylinux1-x64
+        - name: manylinux2014-x86
+        - name: manylinux2014-x64
+        - name: manylinux2014-aarch64
+        - name: manylinux_2_28-x86
+          latest-alias: manylinux-latest-x86
```
azkrishpy, line 53: can we not call it just an "alias" is it reserved keyword?

azkrishpy (commented):
I have not thought through all the cases yet but putting this idea out there.
Can we switch our CIs to a version agnostic naming convention? 
`platform-aarch-min` and `platform-aarch-max` if we support a range within.
`platform-aarch` for a default version we intend to use and test with. I guess this PR kind of takes us in the direction with the idea of aliasing but ideally we just want to update the dockerfile with the same name to use a different version right?

.github/workflows/build-images.yml
sfod: Considering that OS version cannot always be sorted in the right order (AL2023 < AL2), using the explicit manual tags is the most straightforward way. But you're right that it's hard to parse this wall of OSes, so I split it into groups, should be slightly easier to maintain.

.github/workflows/build-images.yml
sfod: `alias` is too generic, no? I don't have a strong opinion on this, so if you think `alias` makes more sense here, I can do renaming.

azkrishpy (approved):

PR #384, opened 2026-09-16
fix(check-abi): diff against the merge base, not the base branch tip

azkrishpy:
**Stack 1 of 3** — independent, reviewable and mergeable on its own.

## The bug
The ABI baseline was `origin/$GITHUB_BASE_REF`, i.e. the **tip** of the base branch. That attributes every change that landed on the base branch since the PR forked to the PR itself. Two consequences:

- a branch behind its base gets a verdict covering commits it never contained;
- the verdict changes as unrelated PRs merge, with no push to the PR.

The push path already fell through to `merge-base`, so the PR path was the *more* wrong of the two.

## The fix
Both branches of the resolution now use `git merge-base HEAD origin/<base-branch>`, taking `<base-branch>` from `GITHUB_BASE_REF` when present and the repository default branch otherwise.

The explicit `base-ref` override still uses the ref **verbatim**: the release path compares "previous tag vs current ref", where the tagged commit itself is the intended baseline, and taking a merge base there would silently compare against something else.

## Validation
Reproduced in a throwaway repo — a branch adding only `g.c`, with `h.c` landing on main afterwards:

```
diff against TIP of main   -> g.c  h.c      # h.c is not this branch's work
diff against MERGE-BASE    -> g.c
```

`bash -n` clean.

## Non-goals
No trigger change, no verdict or labelling change, no new files, no consumer change.

## Rollback
Revert this commit; one file of logic.

---


azkrishpy:
Wrong approach. The caller should resolve the merge base and pass it via check-abi's existing `base-ref` input; the action does not need changing.

PR #385, opened 2026-09-16
fix(check-abi): report semver and source-break as independent axes

azkrishpy:
**Stack 2 of 3** — based on `fix/abi-merge-base` (#384). Review that one first.

## The bug: a source break cannot be unblocked by anyone
Live in every consumer since #378. Three pieces combine:

1. `gate.sh` on a source break set `LABEL=needs-review; REMOVE=""` — needs-review and **no** semver label.
2. `check-abi/action.yml`'s label step drops every verdict label that is not the target, so the PR ends up with **only** `needs-review`.
3. `block-needs-review-label.yml` requires **both** no `needs-review` **and** a `patch`/`minor` label.

So a maintainer who reviews an accepted API break and removes `needs-review` leaves the PR with no semver verdict — and the gate still fails. There is no self-service way out.

Currently latent: nothing ABI-shaped is a required status check in any of the 12 repos, so the gate blocks nothing yet. It stops being latent the moment it becomes required.

## The fix
`gate.sh` emits **two independent axes**:

| case | semver | source break |
|---|---|---|
| clean | `patch` | false |
| binary break, source clean | `minor` | false |
| source break | `minor` | **true** |
| removed enum/macro constants | `minor` | **true** |

A source break reports `minor` rather than suppressing the semver verdict — it is an incompatible API change, so the smallest honest bump is minor. The two labels are then **additive**, and removing `needs-review` leaves a PR that still carries its verdict.

Two marker lines, not one composite token: `action.yml` greps the verdict with `[a-zA-Z0-9_-]*`, which would truncate `minor+needs-review` at the `+` and silently drop the second axis.

## Validation, executed against the real `gate.sh` with fixtures
```
clean              -> ABI_LABEL_RESULT::patch  ABI_SOURCE_BREAK_RESULT::false
binary break       -> ABI_LABEL_RESULT::minor  ABI_SOURCE_BREAK_RESULT::false
source break       -> ABI_LABEL_RESULT::minor  ABI_SOURCE_BREAK_RESULT::true
removed constants  -> ABI_LABEL_RESULT::minor  ABI_SOURCE_BREAK_RESULT::true
tool error (rc=2)  -> exit 2, no markers emitted
```

And against the gate's own logic, with a stubbed `gh`:
```
BEFORE  {needs-review}        rc=1
        {}                    rc=1   <- the deadlock: reviewer removed the label, still blocked
AFTER   {minor,needs-review}  rc=1   <- blocked, as intended
        {minor}               rc=0   <- reviewer removes needs-review, UNBLOCKED
        {patch}               rc=0
```

## Knock-on change
`label` can no longer return `needs-review`, so `compute-version-bump.sh`'s rule 1 keys off the new `source-break` output instead of the label string. A missing second marker reads as `false`, so an older `gate.sh` degrades to today's semver-only behaviour rather than misfiring.

## Non-goals
No trigger change. No change to the gate workflow's own conditions — it is left requiring `patch|minor`, which is now satisfiable. No consumer change.

## Rollback
Revert; restores today's behaviour, deadlock included.

---


azkrishpy:
Wrong approach. Modifies the existing action. The deadlock belongs in the new Pre-merge checks workflow's own labelling, which can apply the semver label alongside needs-review.

PR #387, opened 2026-09-16
feat(changelog): add the changelog action

azkrishpy:
**Stack 4 of 4** — based on `feat/pre-merge-checks` (#386). Adds files only; **nothing references this action on merge.**

Completes the set the stack set out to deliver: ABI labelling, the label/release-impact check, and the changelog check, all reachable from one `Pre-merge checks` workflow.

## What this is
A fragment-based changelog. Authors add `.changes/preview/<PR>.json` alongside their PR; CI checks it; `CHANGELOG.md` is always regenerated from the fragments rather than hand-edited. The PR-time `check` half is consumed by the changelog job added in #386, which stays **off by default** — turning it on is a per-repo `changelog-enabled: true`, so the 2.8% historical PR-title pass rate does not land inside a trigger migration.

## Relationship to #379
This supersedes #379, which is 1 commit on an old base. **#379 is left open and untouched** — closing it is your call, not mine. Carried over from it:

- `scripts/new-change` — interactive helper so an author creates a fragment locally instead of hand-writing JSON. Only existed on that branch; salvaged here.

Deliberately **not** carried over, both with reasons:

- `examples/changelog-rollup.yml` — obsoleted by folding the rollup into `cut-release.sh`, so `VERSION` and `CHANGELOG.md` land in one commit and cannot disagree about what shipped. **That integration is not in this PR** — it arms a `CHANGELOG.md` rewrite in 12 repos pushed straight to `main` by the release deploy key, so it needs its own change behind an opt-in input.
- `examples/changelog-check.yml` — superseded by #386's changelog job. The example built its file list with `git diff ... || true`, so any git failure produced an empty list and **waived** the fragment requirement. Under `pull_request_target`, with no head checkout, that would have been every fork PR. #386's job fails closed instead.

## README corrections
The Modes table listed all eight `changelog.py` subcommands without noting that `seed`, `rollup` and `freeze-snapshot` are CLI-only and that passing them as the action's `mode` exits 2. It also pointed at the deleted check example.

## Validation
71 tests pass (`python3 -m pytest .github/actions/changelog/tests`). Valid YAML; `bash -n` clean on `new-change`.

## Known gaps, called out rather than hidden
These are real and I did not paper over them:

1. **`chore` fragments render nowhere** (`changelog.py:25`, `HIDDEN_TYPES_CUSTOMER`) yet `check` still requires contributors to write them. Either document the exclusion or render them; the current state is the worst of both.
2. **Root `CHANGELOG.md` is not cumulative** — the render walks only `latest/`, so the first minor bump makes the previous line vanish from the root file and survive only in `.changes/<M>.<N>.x/CHANGELOG.md`. Intentional per the README, but worth confirming for 12 GA libraries.
3. **First render overwrites the existing `CHANGELOG.md`.** All 12 repos ship a byte-identical 52-byte file holding `## [1.0.0] / Official release of 1.0.0.`; `render` and `rollup` both `write_text()` over it. Needs a seeding decision before any repo runs the render half.
4. **The `docs`-branch example still uses the cherry-pick-per-merge model**, which has an explicit `Manual intervention required; exit 1` conflict path — so it does not yet meet "docs must never need human interference". The replacement (reset `docs` to `main`, one render commit, force-push with lease) is a separate PR.
5. `test_docs_branch_workflow.py` pins the current cherry-pick model, so it will need rewriting alongside item 4.

## Non-goals
No `cut-release.sh` change. No consumer repo change. `changelog-enabled` stays `false`.

## Rollback
Delete the directory; nothing references it.

---


azkrishpy:
Redundant with #379, which already carries the changelog action. Closing rather than competing with it.

